PR #258, opened 2014-12-02
Fixes #8547 - Showing error when no env params

daviddavis:
To test, try not supplying "--to-lifecycle-environment-id" and "--to-lifecycle-environment" and you should get an error message:

```
Could not promote the content view:
  Error: Missing options to search lifecycle_environment
```


komidore64:
do these come from the server now?

daviddavis:
No, they're not used for anything. To promote, you just need an environment (where the content view version is going to) and the version info.

komidore64:
cool

ack

PR #260, opened 2014-12-04
Fixes #8583 - Renaming id attribute to avoid lookup

daviddavis:


lib/hammer_cli_katello/content_view_puppet_module.rb
```diff
@@ -29,11 +29,8 @@ def extend_data(mod)
     class CreateCommand < HammerCLIKatello::CreateCommand
       command_name "add"
 
-      option "--id", "ID", _("id of the puppet module to associate")
-
-      def request_params
-        super.tap { |params| params['uuid'] = params.delete('id') if params['id'] }
-      end
+      option "--id", "ID", _("id of the puppet module to associate"),
+        :attribute_name => "uuid"
```
komidore64: and this gets put into the request params?

daviddavis: Yea, it's a parameter in the API.

komidore64: awesome

komidore64:
ack

PR #268, opened 2015-01-09
fixes #8892 - requires label/value to set content-override, BZ1173723

cfouant:


lib/hammer_cli_katello/activation_key.rb
```diff
@@ -234,6 +234,10 @@ class ContentOverrideCommand < HammerCLIKatello::SingleResourceCommand
       desc _("Override product content defaults")
       command_name "content-override"
 
+      validate_options do
+        all(:option_label, :option_value).required
```
komidore64: this validation can't be added server-side?

cfouant: It's not required on the UI side, they pass in a hash param, while cli passes in these individual values... so if it were set to required, it would fail any UI updates.

komidore64: if it's not required in the UI, why should it be required here? i'm not sure i follow.

cfouant:
fixed server-side

PR #269, opened 2015-01-13
fixes #8913 - add back content view options for content view version promote

jlsherrill:
after apipie changes made content_view_id optional

### DIFF
--- a/lib/hammer_cli_katello/id_resolver.rb
+++ b/lib/hammer_cli_katello/id_resolver.rb
@@ -61,6 +61,7 @@ module HammerCLIKatello
     def content_view_version_id(options)
       key_id = HammerCLI.option_accessor_name("id")
       key_environment_id = HammerCLI.option_accessor_name("environment_id")
+      key_content_view_id = HammerCLI.option_accessor_name("content_view_id")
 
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/id_resolver.rb
```diff
@@ -72,6 +73,16 @@ def content_view_version_id(options)
...
       end
+
+      begin
+        options[key_content_view_id] ||= content_view_id(
+          scoped_options("content_view", options)
+        )
+      rescue HammerCLIForeman::MissingSeachOptions # rubocop:disable HandleExceptions
+        # Intentionally suppressing the exception,
+        # content_view is not always required.
+      end
```
komidore64: this `L81-84` block feels a little fishy. is there a way you could accomplish this without using `begin...rescue`? just feels dirty.

jlsherrill: Without changing HammerCLIForeman's id_resolver, i don't believe so.  This is used in other places.

komidore64: ah, okay.

komidore64:
looks sane to me! ack

PR #275, opened 2015-02-04
Fixes #8203, #9226 - Docker counts displayed on repo info

parthaa:
Docker Images / Docker Tag information can be now
seen in the hammer repository info command

### DIFF
--- a/lib/hammer_cli_katello/repository.rb
+++ b/lib/hammer_cli_katello/repository.rb
@@ -33,6 +33,8 @@ module HammerCLIKatello
         field :url, _("URL")
         field :_publish_via_http, _("Publish Via HTTP")
         field :full_path, _("Published At")
+        field :docker_upstream_name, _("Upstream Repository Name"),
+              Fields::Field, :hide_blank => true
 
         label _("Product") do
           from :product do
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/repository.rb
```diff
@@ -33,7 +33,8 @@ class InfoCommand < HammerCLIKatello::InfoCommand
         field :url, _("URL")
         field :_publish_via_http, _("Publish Via HTTP")
         field :full_path, _("Published At")
-
+        field :docker_upstream_name, _("Upstream Repository Name"),
+                                      Fields::Field, :hide_blank => true
```
komidore64: why is this indented way over?

daviddavis: :+1 I think if we upgrade rubocop, it'll catch stuff like this.

komidore64: well also (more so in katello) we have a lot of cops simply turned of instead of changed their options. with rubocop `0.28.0`, a lot of cops are configurable.

daviddavis: @komidore64 I officially nominate you to standardize rubocop versions and configs across all projects.

komidore64: i accept this nomination

komidore64:
ack

PR #280, opened 2015-02-28
Cannot use ``none`` for Sync Plan interval.

omaciel:
If you pass ``none`` as the interval for a Sync Plan you get an error.
Turns out not even the API or UI allow you to create a sync plan with no
interval set.

I have updated the code to no longer list ``none`` as a valid option and
set ``weekly`` as the new default value.

Also, a default value is no longer provided for the ``sync_date`` field.

### DIFF
--- a/lib/hammer_cli_katello/sync_plan.rb
+++ b/lib/hammer_cli_katello/sync_plan.rb
@@ -28,15 +28,13 @@ module HammerCLIKatello
     class CreateCommand < HammerCLIKatello::CreateCommand
 
       option "--interval", "INTERVAL", _("how often synchronization should run"),
-             :default => 'none',
              :format => HammerCLI::Options::Normalizers::Enum.new(
-                %w(none hourly daily weekly)
+                %w(hourly daily weekly)
               )
 
       option "--sync-date", "SYNC_DATE",
              _("start date and time of the synchronization defaults to now"),
-             :format => HammerCLI::Options::Normalizers::DateTime.new,
-             :default => DateTime.now.strftime("%F %T")
+             :format => HammerCLI::Options::Normalizers::DateTime.new
 
       success_message _("Sync plan created")
       failure_message _("Could not create the sync plan")
@@ -48,7 +46,7 @@ module HammerCLIKatello
 
       option "--interval", "INTERVAL", _("how often synchronization should run"),
              :format => HammerCLI::Options::Normalizers::Enum.new(
-                %w(none hourly daily weekly)
+                %w(hourly daily weekly)
               )
       option "--sync-date", "SYNC_DATE", _("start date and time of the synchronization"),
              :format => HammerCLI::Options::Normalizers::DateTime.new


omaciel:
Current behavior:

```bash
[root@qe-foreman-rhel66 ~]# hammer -u admin -p changeme sync-plan create -h
Usage:
    hammer sync-plan create [OPTIONS]

Options:
 --description DESCRIPTION               sync plan description
 --enabled ENABLED                       enables or disables synchronization
                                         One of true/false, yes/no, 1/0.
 --interval INTERVAL                     how often synchronization should run
                                         One of 'none', 'hourly', 'daily', 'weekly'
                                         Default: "none"
 --name NAME                             sync plan name
 --organization ORGANIZATION_NAME        Organization name to search by
 --organization-id ORGANIZATION_ID       organization ID
 --organization-label ORGANIZATION_LABEL Organization label to search by
 --sync-date SYNC_DATE                   start date and time of the synchronization defaults to now
                                         Date and time in YYYY-MM-DD HH:MM:SS or ISO 8601 format
                                         Default: "2015-02-28 14:27:02"
 -h, --help                              print help

[root@qe-foreman-rhel66 ~]# hammer -u admin -p changeme sync-plan create --name foobarbaz --organization-id 1 --enabled=False
Could not create the sync plan:
  Validation failed: Interval is not included in the list

[root@qe-foreman-rhel66 ~]# hammer -u admin -p changeme sync-plan create --name foobarbaz --organization-id 1 --enabled=False --interval='none'
Could not create the sync plan:
  Validation failed: Interval is not included in the list
```

omaciel:
Proposed change would provide:

```bash
[root@qe-foreman-rhel66 ~]# hammer -u admin -p changeme sync-plan create -h
Usage:
    hammer sync-plan create [OPTIONS]

Options:
 --description DESCRIPTION               sync plan description
 --enabled ENABLED                       enables or disables synchronization
                                         One of true/false, yes/no, 1/0.
 --interval INTERVAL                     how often synchronization should run
                                         One of 'hourly', 'daily', 'weekly'
                                         Default: "weekly"
 --name NAME                             sync plan name
 --organization ORGANIZATION_NAME        Organization name to search by
 --organization-id ORGANIZATION_ID       organization ID
 --organization-label ORGANIZATION_LABEL Organization label to search by
 --sync-date SYNC_DATE                   start date and time of the synchronization defaults to now
                                         Date and time in YYYY-MM-DD HH:MM:SS or ISO 8601 format
                                         Default: "2015-02-28 14:31:03"
 -h, --help                              print help

[root@qe-foreman-rhel66 ~]# hammer -u admin -p changeme sync-plan create --name 'brand new' --organization-id 1 --enabled false
Sync plan created

[root@qe-foreman-rhel66 ~]# hammer -u admin -p changeme sync-plan info --id 33
ID:          33
Name:        brand new
Start Date:  2015/02/28 14:30:52
Interval:    weekly
Description:
Created at:  2015/02/28 19:30:52
Updated at:  2015/02/28 19:30:52
```

komidore64:
are there any other commands that have the `--interval` option?

komidore64:
i'm thinking we should maybe break this option out into a module

komidore64:
@tstrachota what does passing `none` do? from 7c038691d9cb15b2ac0f7a3ec7793dc6c1e20172

komidore64:
@omaciel, you are correct. looks like the API doesn't accept `none` at all.

apidoc:
```json
{
  "name": "interval",
  "full_name": "interval",
  "description": "\n<p>how often synchronization should run</p>\n",
  "required": false,
  "allow_nil": false,
  "validator": "Must be one of: hourly, daily, weekly.",
  "expected_type": "string",
  "metadata": null,
  "show": true
}
```

Ichimonji10:
@komidore64 The API does not accept `nil` as an interval.

komidore64:
since the server requires `interval` on create, shouldn't we require it here instead of picking for the user?

omaciel:
@komidore64 if we force the user to provide the interval, which sounds like the right thing to do imho, then we should also require that the ``sync_date`` be provided (we currently provide a default value)?

komidore64:
ah. yeah okay. i'd say this is bug worthy (because this PR needs to be linked to a redmine issue anyways) something like "don't provide default sync_plan values for required fields".

omaciel:
@komidore64 are you saying I should create another PR (or re-work this one?) to remove the code that provides defaults?

komidore64:
altering this PR is fine. yeah, I'm in favor of removing the default values for fields that are required server-side. the CLI doesn't (that i'm aware of) do that anywhere else. thanks @omaciel!

omaciel:
PR updated

omaciel:
@komidore64 ping?

komidore64:
@omaciel sorry! didn't see you updated... 

format the commit message like the following and then you're good to go.

```
fixes #<redmine issue number> - <message message message>, BZ<bz number>
```

ack

omaciel:
@komidore64 no worries but... there's no BZ or Redmine issue... this is one of those "heck, this is not right... let me fix it" kind of fixes. What do I do?

komidore64:
@omaciel would you create a redmine issue? we try to have all code changes tied to redmine issues at a minimum, and corresponding BZs if those exist.

omaciel:
@komidore64 filed Redmine issue and changed commit message.

komidore64:
ack

PR #284, opened 2015-03-09
fixes #9679 - fix host collection erratum install

stbenjam:
Errata should be renamed to content, like packages is here:
  https://github.com/Katello/hammer-cli-katello/blob/master/lib/hammer_cli_katello/host_collection_package.rb#L13


daviddavis:
ACK

PR #290, opened 2015-04-02
Fixes #9995: Only send content IDs that user passes during incremental u...

ehelms:
...pdate.

### DIFF
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -152,6 +152,13 @@ module HammerCLIKatello
           params.delete('update_systems')
         end
 
+        add_content = {}
+        params['add_content'].each do |key, value|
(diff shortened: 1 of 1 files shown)


lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -152,6 +152,13 @@ def request_params
           params.delete('update_systems')
         end
 
+        add_content = {}
+        params['add_content'].each do |key, value|
+          add_content[key] = value if options.key?("option_#{key}")
```
komidore64: small correction: use `HammerCLI.option_accessor_name("foo")` you want to pull something from `options`.

so it would look like this:
```ruby
if options.key?(HammerCLI.option_accessor_name(key))
```

komidore64:
ack pending comment

PR #293, opened 2015-05-11
fixes #10473 - do not support prior id on environment update

jlsherrill:


jlsherrill:
Its not actually possible to update the prior, removed from the api docs here:  https://github.com/Katello/katello/pull/5221

ehelms:
APJ

PR #296, opened 2015-05-27
fixes #10623 - cvv resolver finds latest version in --from-lifecycle-environment

dustints:


ehelms:
Reading through the issue, this code appears to implement the first suggestion:

```
hammer content-view version promote --organization Engineering --content-view-id 5 --to-lifecycle-environment DEV
```

Given you can promote any version to any environment, I don't think this would make sense as it is not clear what is getting promoted. The second suggestion would work better:

```
hammer content-view version promote --organization Engineering --content-view-id 5 --to-lifecycle-environment DEV --from-lifecycle-environment Library
```

Given only one version of the view can live in a particular environment, you would at least have 'some' idea of what is being promoted. Question for this second version would be, do we support promoting out of the archive with this syntax and if so, how does the user specify the version to promote?

dustints:
@ehelms updated, it doesn't support the promoting from the archive since it just pick's up the latest version from each env.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -78,6 +78,12 @@ class PromoteCommand < HammerCLIKatello::SingleResourceCommand
              _("Name of the target environment"), :attribute_name => :option_to_environment_name
       option "--to-lifecycle-environment-id", "TO_ENVIRONMENT_ID",
              _("Id of the target environment"), :attribute_name => :option_to_environment_id
+      option "--from-lifecycle-environment", "FROM_ENVIRONMENT_ID",
+             _("Environment from where to promote from"),
+             :attribute_name => :option_from_environment_name
+      option "--from-lifecycle-environment-id", "FROM_ENVIRONMENT_ID",
+             _("Environment from where to promote from"),
```
ehelms: Maybe mention this should be the ID of the environment

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -78,6 +78,12 @@ class PromoteCommand < HammerCLIKatello::SingleResourceCommand
              _("Name of the target environment"), :attribute_name => :option_to_environment_name
       option "--to-lifecycle-environment-id", "TO_ENVIRONMENT_ID",
              _("Id of the target environment"), :attribute_name => :option_to_environment_id
+      option "--from-lifecycle-environment", "FROM_ENVIRONMENT_ID",
+             _("Environment from where to promote from"),
```
ehelms: Maybe mention this should be the name of the environment

dustints:
@ehelms updated descriptions

ehelms:
LGTM, can we squash these commits into 1?

thomasmckay:
closing in favor of https://github.com/Katello/hammer-cli-katello/pull/308

thanks @dustints !

PR #300, opened 2015-06-23
fixes #9876 - Adds override value of product content in activation key, BZ1180282

cfouant:


lib/hammer_cli_katello/activation_key.rb
```diff
@@ -219,11 +219,9 @@ class ProductContentCommand < HammerCLIKatello::ListCommand
         from :content do
           field :id, _("ID")
           field :name, _("Name")
-          field :type, _("Type")
-          field :contentUrl, _("URL")
-          field :gpgUrl, _("GPG Key")
```
ehelms: I understand the adding of the override, but not immediately why remove these at the same time?

cfouant: I was making it uniform to the UI's activation key product content page.

ehelms: Unless those items are completely useless to the user, we should leave them since users may already be used to using them.

cfouant:
@ehelms can you re-review? BZ failing QA until this goes in. (my fault for forgetting to comment dependency)

cfouant:
requires https://github.com/Katello/katello/pull/5133

ehelms:
ACK

PR #302, opened 2015-07-03
Fixes #10062 - Allow users to see repos for a content view

daviddavis:


ehelms:
@daviddavis Does this require the API change first? And is this compatible with master or does it require the OS Tree work first?

daviddavis:
The API change is right above your comment. It should work on master (without any ostree work). It'll show just a blank set for OSTree content. If that's a problem, I can wait until after ostree makes it to master to merge this.

ehelms:
ACK

daviddavis:
Opened new PR: https://github.com/Katello/hammer-cli-katello/pull/310

PR #304, opened 2015-07-24
fixes #8481 - adds content override to hammer content host, BZ1166891

cfouant:


cfouant:
requires https://github.com/Katello/katello/pull/5373

ehelms:
ACK

cfouant:
@ehelms - I updated the command names to have uniformity with requested changes on API PR. If you could re-review when you have opportunity, that would be great. Thanks!

cfouant:
ping @ehelms - could you ACK the changes made? Just needed to update the command names to match API p-r

ehelms:
So I played around with this a little more, and I had a few questions:

1) What do you think about making content-override a sub-command? That might fit better with our patterns so you could do `content-host content-overrides list` and `content-host content-overrides update`?

2) When doing the set-content-override the error messages are a tad confusing:

```
vagrant@katello-bugbox hammer-cli-katello ((c2e6a03...))$ hammer content-host set-content-override --id a8a6823e-6845-448b-af95-6cc79b532aa0 --content-label Default_Organization_Zoo_Zoo5
Could not update content override:
  Missing arguments for 'content_override[value]'
vagrant@katello-bugbox hammer-cli-katello ((c2e6a03...))$ hammer content-host set-content-override --id a8a6823e-6845-448b-af95-6cc79b532aa0 --content-label Default_Organization_Zoo_Zoo5 --value false
Could not update content override:
  Value must be 0/1, or 'default'
```

For example, there is no 'content_override[value]' option, there is a `--value` option. And the values required are 0/1 or default, however, compared to the UI which uses yes and no there is a disconnect.

daviddavis:
@cfouant ping

cfouant:
requires merge of https://github.com/Katello/katello/pull/5853

ehelms:
Now that we have a test setup, what do you think about adding some for these commands?

cfouant:
Per @jlsherrill - content-host is moving over to host, and this will be under host subscriptions. I am closing this PR until that change has merged.

PR #305, opened 2015-07-24
Refs #9848: Move config file to config directory and provide via gem.

ehelms:


ehelms:
Identified this while adding this to katello-packaging and noted that this is the way other hammer modules handle it.

rubygem-hammer_cli_katello.spec
```diff
@@ -52,7 +51,7 @@ gem install --local --install-dir .%{gem_dir} \
 
 %install
 mkdir -p %{buildroot}%{_sysconfdir}/%{confdir}/cli.modules.d
-install -m 755 %{SOURCE1} %{buildroot}%{_sysconfdir}/%{confdir}/cli.modules.d/katello.yml
+install -m 755 .%{gem_instdir}/config/katello.yml %{buildroot}%{_sysconfdir}/%{confdir}/cli.modules.d/katello.yml
```
jlsherrill: i can't actually build the rpm with this change.  For one i think its geminstdir

but even that doesn't get you all the way

jlsherrill: needs a %file entry

jlsherrill: update to use  gem_instdir and gem_name throughout

jlsherrill:
ACK

PR #308, opened 2015-08-06
fixes #10623 - cvv resolver uses max version if > 1 found

thomasmckay:
include from-environment

replacing https://github.com/Katello/hammer-cli-katello/pull/296


thomasmckay:
@ehelms squashed commits

ehelms:
ACK

PR #311, opened 2015-09-03
fixes #8435 - progress bar on repository content-upload

thomasmckay:


ehelms:
Does this introduce powerbar as a new dependency? Or does hammer-cli already provide this? If not, will we need to package this or is it already packaged?

thomasmckay:
@ehelms powerbar is already present and used for async task output.

ehelms:
Oh, from hammer-cli-foreman-tasks ? You'll want to require this in the gemspec just it does. 

There is a question of whether it would be worth creating a module in hammer-cli that encompasses this behavior so that we only have to override what is specific to a given task that wants a progress bar.

thomasmckay:
Added powerbar to the spec. I considered breaking it out into a utility but opted not to since the amount of shared code was small. I figured we could revisit if the pattern came up again.

lib/hammer_cli_katello/repository.rb
```diff
@@ -255,6 +262,27 @@ def import_upload_ids(ids)
         }
         resource.call(:import_uploads, params)
       end
+
+      def progress_message(filename)
+        _('Uploading %{filename}') % {:filename => filename}
+      end
+
+      def render_result
+        print_message _('finished')
```
ehelms: Looks like the trend is capitalized words and what is the effect if errors occur during the upload process? Since the progress bar method is doing an 'ensure' this finished would get printed everytime, but will it be obvious there were errors and what they are?

thomasmckay: Not sure I want to expand this pull-request into more than what's here. Perhaps this was not the "low hanging fruit" issue I had hoped it was. :)

ehelms: I don't think what I am asking is all that much, just to check/ensure that if there is an error state the user is presented with that information.

thomasmckay: Agreed but I'd want to fix it in the other similar code areas as well. I switched this back to work-in-progress.

ehelms: But you didn't introduce this new code in other areas? If there are similar areas, then I agree make changes in a different PR. This one is a nice small atomic change, that what I am asking about should take 15 minutes to verify and adjust if need be.

daviddavis:
what's the status of this?

thomasmckay:
Status: discouraged. Closing.

PR #313, opened 2015-09-09
fixes #11750 - errata apply: update to support sync or async execution

bbuckingham:
Small update to support initiating an errata apply as a sync or
async operation.  This makes the behavior for errata consistent with
the package actions.


ehelms:
ACK

PR #315, opened 2015-09-14
Fixes #11801 : Update the version of gettext, and modify the rake file.

bkearney:
This brings hammer-cli-katello in line with hammer-cli. Also, it would be
good to get this tool into transifex like the other hammer plugins.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ doc/
 .ruby-version
 .ruby-gemset
 .*.sw?
+
+# Locale files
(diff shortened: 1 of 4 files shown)


ehelms:
@bkearney looks like this needs an update

bkearney:
@ehelms fixed

ehelms:
I think you may need to rebase this as I am seeing two commits

bkearney:
@ehelms fixed

ehelms:
ACK

PR #326, opened 2015-10-15
Add 'gem install bundler' to README.md

beav:
This additional command was needed when I ran through the dev setup.


komidore64:
RVM changed what their default gemset includes, methinks.

komidore64:
ACK

PR #327, opened 2015-10-22
fixes #12270 - validates input data for activation key create, BZ1156536

cfouant:


lib/hammer_cli_katello/activation_key.rb
```diff
@@ -81,12 +78,10 @@ class CreateCommand < HammerCLIKatello::CreateCommand
       success_message _("Activation key created")
       failure_message _("Could not create the activation key")
 
-      def request_params
-        params = super
-        if params.key?('max_content_hosts') && !params['max_content_hosts'].match(Common::INTEGER_FORMAT)
```
komidore64: I see that we're removing the `Common::INTEGER_FORMAT`. Is the server going to validate integer format rather than hammer now?

cfouant: The API seems to handle it fine.

komidore64:
ACK

PR #328, opened 2015-10-26
fix class names, use Manifest instead of Manfiest

jmrodri:


ehelms:
Please open a Redmine issue that captures this and update the PR with http://www.katello.org/developers/style/git_commit_messages.html

PR #329, opened 2015-10-29
fixes #12350 - content_view - add copy command

bbuckingham:
This is to add copy command to be consistent with UI behaviors.

### DIFF
--- a/lib/hammer_cli_katello/content_view.rb
+++ b/lib/hammer_cli_katello/content_view.rb
@@ -104,6 +104,19 @@ module HammerCLIKatello
       build_options
     end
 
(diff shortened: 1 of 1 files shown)


komidore64:
this looks good to me.

would someone with a working dev env smoke test this real quick?

ehelms:
This seems to be requiring both `id` and `name` which I assume `name` here is the name of the new content view. However, the help text indicates it is there for searching:

```
 --name NAME                             Content view name to search by
```

bbuckingham:
@ehelms, updated so that the help text now shows: 

```
 --name NAME                             New content view name
```

ehelms:
ACK

PR #336, opened 2015-12-16
Fixes #12846 - Fix hammer-cli-foreman dependency error

johnpmitsch:


mmoll:
Why not use something like `'>= 0.5.0'` or `~> 0.5`? This was changed in hammer-cli-foreman for the reason that this always happens with a new hammer release and semver was designed to take care of such situations.

komidore64:
ack

PR #340, opened 2015-12-17
Fixes #12757: update packages to use host instead of system.

waldenraines:
http://projects.theforeman.org/issues/12757

### DIFF
--- a/lib/hammer_cli_katello.rb
+++ b/lib/hammer_cli_katello.rb
@@ -117,4 +117,7 @@ module HammerCLIKatello
                                          'hammer_cli_katello/erratum'
   )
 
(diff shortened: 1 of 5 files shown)


lib/hammer_cli_katello/output/fields.rb
```diff
@@ -1,3 +1,4 @@
+
```
cfouant: Is there a reason for this?

cfouant:
ACK from me pending the requested nitpick change

jlsherrill:
@waldenraines ping, plz2merge

jlsherrill:
merging, thanks @waldenraines !

PR #343, opened 2016-01-26
Fixes #13397 - Update the zanata mapping file

bkearney:


komidore64:
This looks good. @bkearney is there anything that needs to be done to "test" that this works? I'm assuming since you're pull-requesting that you made sure it worked?

PR #349, opened 2016-02-09
Fixes #13614 - Make capsule only an alias to proxy

tstrachota:


lib/hammer_cli_katello.rb
```diff
@@ -24,6 +24,8 @@ def self.exception_handler_class
   require 'hammer_cli_katello/id_resolver'
 
   # commands
+  require 'hammer_cli_katello/capsule'
```
komidore64: Why can't this be lumped into the list of other requires just a few lines up?

lib/hammer_cli_katello/capsule.rb
```diff
@@ -22,39 +22,9 @@ def all_options
...
-        proxy
-      end
-
-      build_options
-    end
-
-    class Content < HammerCLI::AbstractCommand
+    class Content < HammerCLIKatello::Command
       command_name 'content'
-      desc _('Manage the capsule content')
```
komidore64: I'm in favor of leaving the description set here, so that ``#subcommand()`` down below can reference it. One place to change the description in the future, rather than two.

lib/hammer_cli_katello/capsule.rb
```diff
@@ -136,7 +106,6 @@ class SyncCommand < HammerCLIForemanTasks::AsyncCommand
       autoload_subcommands
     end
 
-    autoload_subcommands
+    HammerCLIForeman::SmartProxy.subcommand "content", _('Manage the smart proxy content'), HammerCLIKatello::Capsule::Content
```
komidore64: nitpick: mind splitting this into three lines?

```ruby
HammerCLIForeman::SmartProxy.subcommand(HammerCLIKatello::Capsule::Content.command_name,
  HammerCLIKatello::Capsule::Content.desc,
  HammerCLIKatello::Capsule::Content)
```

tstrachota:
@komidore64 thanks, updated.

lib/hammer_cli_katello/capsule.rb
```diff
@@ -136,7 +108,9 @@ class SyncCommand < HammerCLIForemanTasks::AsyncCommand
       autoload_subcommands
     end
 
-    autoload_subcommands
+    HammerCLIForeman::SmartProxy.subcommand("content",
```
komidore64: One more thing. Could we use ``HammerCLIKatello::Capsule::Content.command_name`` here rather than ``"content"``? Same reason that I asked about using ``#desc()``, it's one place to keep track of the command name, rather than two.

komidore64:
ACK pending [comment](https://github.com/Katello/hammer-cli-katello/pull/349#discussion-diff-53806796).

tstrachota:
Fixed

komidore64:
@tstrachota mind rebasing? Also, is this intended for 3.0? We branched yesterday and this will need a BZ attached to it, if you want it cherry-picked into 3.0.

tstrachota:
@komidore64 I rebased, moved the issue to Katello and cloned it to bugzilla.

komidore64:
@tstrachota sweet! :cat2:

PR #353, opened 2016-02-19
Fixes #13615 and #13616 - capsule sync status and cancel

tstrachota:
Contains:
- command for canceling capsule synchronizations:
   - `capsule content cancel-synchronization`
- 2 commands for capsule status:
    - `capsule content synchronization-status` to list running tasks
    - `capsule content info` to get counts of actually synced content
- base for testing katello commands (requires https://github.com/theforeman/hammer-cli/pull/194 and https://github.com/theforeman/hammer-cli-foreman/pull/215)
- functional tests for the new commands


test/test_helper.rb
```diff
@@ -0,0 +1,16 @@
+require 'minitest/autorun'
+require 'minitest/spec'
+require "mocha/setup"
+
+require 'hammer_cli'
+require 'hammer_cli_foreman/commands'
+
+KATELLO_VERSION = Gem::Version.new(ENV['TEST_API_VERSION'] || '1.5')
```
mbacovsky: Isn't `2.5` the current version?

tstrachota: Indeed, it is! Fixed.

komidore64:
Do we use the full word ``synchronization`` in hammer? Or usually just ``sync``? Let me confirm.

lib/hammer_cli_katello/capsule.rb
```diff
@@ -133,6 +133,105 @@ class SyncCommand < HammerCLIForemanTasks::AsyncCommand
...
+
+        build_options
+      end
+
+      class InfoCommand < HammerCLIKatello::InfoCommand
+        resource :capsule_content, :sync_status
+        command_name "info"
+
+        output do
+          collection :lifecycle_environments, _("Environments") do
```
komidore64: I think this title should be ``Lifecycle Environments`` to not confuse with puppet environments?

lib/hammer_cli_katello/capsule.rb
```diff
@@ -133,6 +133,105 @@ class SyncCommand < HammerCLIForemanTasks::AsyncCommand
...
+                  field :package_groups, _('Package groups')
+                  field :errata, _('Errata')
+                  field :puppet_modules, _('Puppet modules')
+                end
+              end
+            end
+          end
+        end
+
+        def extend_data(data)
```
komidore64: Is overriding this method necessary since you aren't manipulating the data in any way?

lib/hammer_cli_katello/output/fields.rb
```diff
@@ -10,6 +10,9 @@ class ChecksumFilePair < ::Fields::Field
       class Dependency < ::Fields::Field
       end
 
+      class Array < ::Fields::Field
```
komidore64: How does this differ from ``Fields::List``?

mbacovsky:
As for tests part it looks good and is great and long missing addition. It would be good to change the Rakefile to run the functional tests as well or use travis config to let Travis run the tests.

lib/hammer_cli_katello/capsule.rb
tstrachota: I forgot to delete it. It's removed now.

lib/hammer_cli_katello/output/fields.rb
tstrachota: The difference is in how it's formatted https://github.com/Katello/hammer-cli-katello/pull/353/files#diff-9ae5ae9df1629bf9fc24b04247dabfcdR29
`List` is printed on one line, items separated by commas, while `Array` puts each item on it's own line. So it's a cosmetic thing. We could probably just keep both as list and switch with a field parameter too.

komidore64: Yeah, I was thinking similar. Keeping list and adding an optional ``separator`` argument. Or possibly extending ``List`` with ``VerticalList`` only changing the separator that ``List`` uses?

tstrachota: That makes sense. It will require a PR into hammer-cli though. I'll attach a link once the PR is ready.

tstrachota:
@komidore64 https://github.com/theforeman/hammer-cli/pull/195 contains the formatter changes.

lib/hammer_cli_katello/capsule.rb
```diff
@@ -133,6 +133,101 @@ class SyncCommand < HammerCLIForemanTasks::AsyncCommand
...
+          field :_status, _('Status')
+          collection :active_sync_tasks, _('Currently running sync tasks') do
+            field :id, _('Task id')
+            field :_progress, _('Progress')
+          end
+          label _('Last failure') do
+            from :_last_failure do
+              field :id, _('Task id')
+              from :humanized do
+                field :errors, _('Messages'), Fields::List, :on_new_line => true, :separator => "\n"
```
komidore64: :fire: :fire: :fire: :+1:

komidore64:
ACK pending rebase!

ehelms:
@tstrachota the two issues indicated here, given they are being fixed in hammer-cli-katello look like they should be moved over to the Katello project

tstrachota:
@ehelms @komidore64 I rebased, moved the issues to Katello and cloned them to bugzilla.

tstrachota:
One more thing: I raised dependency on hammer-cli-foreman to 0.6.0 as we need the field changes and tests from there.

tstrachota:
@komidore64 is there anything else blocking this PR from being merged?

komidore64:
thanks @tstrachota :beers:

PR #354, opened 2016-02-24
Fixes #13875 - Fixing the prior option in the CLI

daviddavis:


ehelms:
Fixes the bats tests for me, ACK!

PR #359, opened 2016-03-07
Fixes #14065 - call correct command in capsule content test

tstrachota:
One of the tests called `hammer proxy content list` instead of `hammer proxy content info` by mistake.

```
  1) Failure:
capsule content info#test_0003_is mounted under proxy too [/hammer-cli-katello/test/functional/capsule/content/info_test.rb:60]:
The exit code was expected to be EX_OK (0), but it was EX_USAGE (64)
```


ehelms:
ACK - thanks @tstrachota !

PR #360, opened 2016-03-07
Fixes #14068 - Allow creating content views with repository names

adamruzicka:


ehelms:
Please squash into a single commit.

komidore64:
i'm not sure what rubocop is doing...

bbuckingham:
@adamruzicka, If possible, please go ahead and rebase with master.  I believe the following PR recently merged will address the rubocop failure: https://github.com/Katello/hammer-cli-katello/pull/361 .  Thanks!

adamruzicka:
@komidore64, @bbuckingham: Thanks for looking into this. I did a rebase, let's hope it gets fixed

bbuckingham:
@adamruzicka, Thanks! The fix tested out well for me.  ACK.

@ehelms, @komidore64, any additional comments before merging?

komidore64:
:expressionless: Looks like another rebase is needed.

No additional comments from me.

ehelms:
Now that we have a test setup, what do you think about adding some for these updates?

adamruzicka:
@ehelms Added some tests which should cover the updates

lib/hammer_cli_katello/id_resolver.rb
```diff
@@ -61,6 +62,18 @@ def repository_id(options)
...
+    def repository_ids(options)
+      return options['option_repository_ids'] unless options['option_repository_ids'].nil?
+
+      key_names = HammerCLI.option_accessor_name 'names'
+      key_product_id = HammerCLI.option_accessor_name 'product_id'
+      options[key_product_id] ||= product_id(scoped_options 'product', options)
+      find_resources(:repositories, options)
+        .select { |repo| options[key_names].include? repo['name'] }.map { |repo| repo['id'] }
+      # ^ This could be removed if Katello's repositories controller allowed to filter
+      #   repositories index by repository names
```
ehelms: I tend towards not including this comment, this is a general issue with how tenancy works in Katello and I don't want someone reading this thinking it is a simple change to the repositories controller. Right now filtering by name can occur via scoped_search or if we combined name + product to scope things properly.

ehelms:
Thanks for adding tests! @tstrachota or @mbacovsky would you mind taking a look at the tests briefly? Since we haven't had tests in here before, I am still getting up to speed on what a good test within hammer should look like.

lib/hammer_cli_katello/id_resolver.rb
adamruzicka: I see, removed

test/functional/content-view/create_test.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    api_expects(:content_views, :create, 'Create content view') do |par|
+      par['name'] == name &&
+        par['repository_ids'] == ids &&
+        par['organization_id'] == org_id
+    end
+
+    expected_result = success_result("Content view created\n")
+    result = run_cmd(@cmd + @base_params + params)
+    sanitize.call(result)
```
tstrachota: If the warning is expected behavior, I'd prefer changing the stderr expectation over clearing the result. 
```ruby
expected_result = success_result("Content view created\n")
expected_result.expected_err = 'Warning: ...'
result = run_cmd(@cmd + @base_params + params)
assert_cmd(expected_result, result)
```

tstrachota:
@ehelms @adamruzicka One note in the tests. Otherwise it looks good!

test/functional/content-view/create_test.rb
adamruzicka: Created issues for the deprecation warning problem[1] and test refactoring[2]

[1] - http://projects.theforeman.org/issues/14289
[2] - http://projects.theforeman.org/issues/14290

tstrachota: Just for the record:
After a further investigation it turned out the printed warning is a bug. The warning should be displayed only when users really pass `--id`. Currently it occurs every time the content-view command is loaded.

tstrachota: One more idea:
```ruby
require 'hammer_cli_katello/content_view'
```
is probably cleaner workaround than deleting stderr.

adamruzicka: Incorporated that into the latest commit, works like a charm. Thanks!

adamruzicka:
@bbuckingham @ehelms @komidore64 any more comments on this?

komidore64:
None from me.

ehelms:
Nothing from me -- looks like it needs a rebase

adamruzicka:
Rebased and the tests are green

ehelms:
ACK - thanks @adamruzicka !

PR #362, opened 2016-03-10
Fixes #14147 - Fix content_view_filter create to accept repository names

akofink:
It seems the tests weren't catching this bug. Perhaps we should implement tests with this PR before merging.

### DIFF
--- a/lib/hammer_cli_katello/activation_key.rb
+++ b/lib/hammer_cli_katello/activation_key.rb
@@ -10,7 +10,7 @@ module HammerCLIKatello
       output do
         field :id, _("ID")
         field :name, _("Name")
-        field :format_consumed, _("Consumed")
+        field :format_consumed, _("Host Limit")
         from :environment do
           field :name, _("Lifecycle Environment")
         end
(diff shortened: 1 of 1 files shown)


bbuckingham:
@akofink, https://github.com/Katello/hammer-cli-katello/pull/361 was merged to master; therefore, you should be able to rebase with master to get @ehelms' commit that fixed the initial rubocop error.  Also, you will want to squash all commits to 1 before the PR can be merged to master.

akofink:
I'm not sure what's going on here, but I may need help fixing it.

ehelms:
I see you added a test for one of the functions, what about the rest?

akofink:
@ehelms I can add those tests, but I didn't modify those methods. Because I modified `create_repositories_search_options`, I added a test for that method in this PR. We may want to create an issue for making the rest of the tests in this project.

ehelms:
I feel like history has shown all encompassing issues like what you suggest rarely go anywhere because they are too encompassing. Requiring tests where appropriate from here on out will get us there faster I believe because it involves small incremental changes.

akofink:
I agree. So you were joking in your first comment, then?

PR #364, opened 2016-03-14
fixes #14174 - content-view add-repository should require user to provide repo info

bbuckingham:
This commit addresses the issue that if the user was attempting to add
a repository to a content view, but did not specify the repository parameters,
the command would succeed indicating "The repository has been associated".

The update behavior is:

hammer> content-view add-repository --name zoo2 --organization-id 3
Could not add repository:
  Error: At least one of options --repository-id, --repository is required

### DIFF
--- a/lib/hammer_cli_katello/associating_commands.rb
+++ b/lib/hammer_cli_katello/associating_commands.rb
@@ -28,6 +28,11 @@ module HammerCLIKatello
         command_name 'add-repository'
         associated_resource :repositories
 
+        def validate_options
+          super
+          validator.any(:option_repository_id, :option_repository_name).required
+        end
+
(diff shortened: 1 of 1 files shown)


komidore64:
Should this validation be done by the server? Or no?

bbuckingham:
@komidore64, in this particular case, we cannot perform the validation server-side as we are using the 'update' action on the content views controller to add/remove repositories.  This API/action can be used for updating many attributes of the content view; therefore, repository updates are optional.

komidore64:
@bbuckingham ah, okay. I'm cool with that then.

ACK

bbuckingham:
@komidore64, Cool.  Thanks for the review!

PR #365, opened 2016-03-15
fixes #14203 - host-collection create: generating 'option_host_ids' error

bbuckingham:
This commit is to address an error during creation of a host collection.

Before:
-------
hammer> host-collection create --organization-id 1 --name cli-test-2
Could not create the host collection:
  Error: undefined local variable or method `option_host_ids' for #<HammerCLIKatello::HostCollection::CreateCommand:0x00000005567958>

After:
------
hammer> host-collection create --organization-id 1 --name cli-test-2
Host collection created


komidore64:
ACK

PR #367, opened 2016-03-18
Fixes #14147 - Allow repo names in creating content view filters

akofink:
Continuation of #362

### DIFF
--- a/lib/hammer_cli_katello/id_resolver.rb
+++ b/lib/hammer_cli_katello/id_resolver.rb
@@ -1,3 +1,5 @@
+require 'hammer_cli_katello/search_options_creators'
+
 module HammerCLIKatello
 
   class Searchables < HammerCLIForeman::Searchables
@@ -39,6 +41,11 @@ module HammerCLIKatello
 
   # rubocop:disable ClassLength
   class IdResolver < HammerCLIForeman::IdResolver
+    include HammerCLIKatello::SearchOptionsCreators
+
+    # alias_method_chain :create_search_options, :katello_api
+    alias_method :create_search_options_without_katello_api, :create_search_options
+    alias_method :create_search_options, :create_search_options_with_katello_api
 
     def capsule_content_id(options)
       smart_proxy_id(options)
@@ -80,7 +87,9 @@ module HammerCLIKatello
 
       key_names = HammerCLI.option_accessor_name 'names'
       key_product_id = HammerCLI.option_accessor_name 'product_id'
-      options[key_product_id] ||= product_id(scoped_options 'product', options)
+      unless options['option_product_name'].nil?
+        options[key_product_id] ||= product_id(scoped_options 'product', options)
+      end
       find_resources(:repositories, options)
         .select { |repo| options[key_names].include? repo['name'] }.map { |repo| repo['id'] }
     end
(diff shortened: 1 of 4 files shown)


akofink:
@ehelms Any more tests that might be relevant to these changes?

lib/hammer_cli_katello/id_resolver.rb
```diff
@@ -38,6 +38,14 @@ def for(resource)
   end
 
   class IdResolver < HammerCLIForeman::IdResolver
+    require 'hammer_cli_katello/search_options_creators'
```
ehelms: This seem strange to need to do here in general and to do it at least at the top of the file if required.

lib/hammer_cli_katello/id_resolver.rb
```diff
@@ -38,6 +38,14 @@ def for(resource)
...
+    include HammerCLIKatello::SearchOptionsCreators
+
+    # alias_method_chain :create_search_options, :katello_api
+    alias_method :create_search_options_without_katello_api,
+      :create_search_options
+    alias_method :create_search_options,
+      :create_search_options_with_katello_api
```
ehelms: Did these two break the 120 character mark?

ehelms:
Tests look great!

jlsherrill:
@akofink this doesn't appear to fix the original issue:

```
hammer content-view  filter create --content-view=test  --organization="Default Organization"  --repositories=test --type rpm  --name=foo3 
```

seemed to select more than just the 'test' repo when creating the filter

akofink:
Merge with https://github.com/Katello/katello/pull/5950

akofink:
@ehelms The original issue should be fixed now.

komidore64:
@akofink did rubocop yell about a class being too large? Is that why you split out ``search_options_creators.rb``?

Does it make more sense to have one ``search_options_creators.rb`` file for all of them or a ``<thing>_options_creators.rb`` for each creator? That way we could keep the modules small? Or is that not worth it?

akofink:
@tstrachota Yes, I admit I separated it for rubocop.

As for splitting into smaller modules, there is common functionality for a few of the creators, such as [L39](https://github.com/akofink/hammer-cli-katello/blob/14147/lib/hammer_cli_katello/search_options_creators.rb#L39),[L43](https://github.com/akofink/hammer-cli-katello/blob/14147/lib/hammer_cli_katello/search_options_creators.rb#L43),[L47](https://github.com/akofink/hammer-cli-katello/blob/14147/lib/hammer_cli_katello/search_options_creators.rb#L47),[L51](https://github.com/akofink/hammer-cli-katello/blob/14147/lib/hammer_cli_katello/search_options_creators.rb#L51). I think this shared logic would complicate the effort of splitting them into smaller modules, though I'm open to discussion of course.

komidore64:
One file is good for now. I'm was just curious if that was why.

Also: it looks like you'll need to rebase with upstream.

ACK.

akofink:
@komidore64 rebased.

jlsherrill:
Seems to resolve the original issue, ACK from a functionality perspective

jlsherrill:
Merging  based on my ACK and @komidore64 ACK  Thanks @akofink

PR #370, opened 2016-03-24
Fixes #14333 - Makes --unlimited-hosts a flag for update on an activation key.

zjhuntin:
This will make specifying unlimited-hosts in the update command the same as specifying it in the create command. It shoud be a flag because there would never be a time where someone would specify      --unlimited-hosts= false.


cfouant:
ACK - works great thanks @zjhuntin

bbuckingham:
Merging based upon the ACK.  Thanks @zjhuntin !

PR #372, opened 2016-03-28
fixes #14365 - host-collection create|update - fix to support host input

bbuckingham:
Prior to host-unification, users could provide a list of content hosts
as inputs to the host-collection create and update commands.  This
commit should now allow them to provide a list of 'hosts'.

The following are a few valid scenarios:

hammer> host-collection create --organization-id 1 --name with-hosts --host-ids 1
Host collection created

hammer> host-collection create --organization-id 1 --name with-hosts-by-name --hosts "rhel7-client.test"
Host collection created

hammer> host-collection update --id 4 --host-ids 1
Host collection updated

hammer> host-collection update --id 5 --hosts 'rhel7-client.test'
Host collection updated

### DIFF
--- a/lib/hammer_cli_katello/host_collection.rb
+++ b/lib/hammer_cli_katello/host_collection.rb
@@ -3,22 +3,6 @@ module HammerCLIKatello
   class HostCollection < HammerCLIKatello::Command
     resource :host_collections
 
-    module UuidRequestable
-      def self.included(base)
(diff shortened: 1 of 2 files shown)


komidore64:
@bbuckingham ytm

ACK

bbuckingham:
@komidore64 , Thanks!

PR #373, opened 2016-03-30
Fixes #14397 - Do not require a authentication for `hammer ping`

akofink:


test/functional/ping_test.rb
```diff
@@ -0,0 +1,11 @@
+require_relative 'test_helper'
+
+describe 'ping' do
+  before(:each) do
+    HammerCLIKatello::Command.any_instance.stubs(:send_request).returns('services' => [])
```
tstrachota: Have you considered using `api_expects` construct to stub the api call? Something like: https://github.com/Katello/hammer-cli-katello/blob/master/test/functional/host/subscription/unregster_test.rb#L13-L16
It will enable for checking whether the command hits the correct api endpoint.

akofink:
Might want to wait until #374 is merged so the tests can be run.

akofink:
@tstrachota Thanks for the recommendation!

akofink:
Ping @Katello/cli this should be ready to merge.

komidore64:
ACK

PR #378, opened 2016-04-07
Fixes #14538 - Resolver will not try to find generic lifecycle_environment if not in options.

zjhuntin:
To test try to promote a content-view version with a to and from lifecycle and include the organization-id.


bbuckingham:
It appears that the following commands have a similar issue to the one that this PR is addressing.  Would it be possible to also fix them as part of this PR?

```
hammer> content-view list --organization-id 1
Error: found more than one lifecycle_environment

hammer> activation-key list --organization-id 1
Error: found more than one lifecycle_environment
```

lib/hammer_cli_katello/id_resolver.rb
```diff
@@ -93,23 +93,17 @@ def content_view_version_id(options)
...
 
-      begin
-        options[key_environment_id] ||= lifecycle_environment_id(
-          scoped_options("environment", options)
-        )
-      rescue HammerCLIForeman::MissingSeachOptions # rubocop:disable HandleExceptions
+      if options[key_environment_id]
+        options[key_environment_id] ||= search_and_rescue(
+          :lifecycle_environment_id, "environment", options)
         # Intentionally suppressing the exception,
```
bbuckingham: since the exception handling is now in search_and_rescue, we can probably move this comment there

lib/hammer_cli_katello/id_resolver.rb
```diff
@@ -93,23 +93,17 @@ def content_view_version_id(options)
...
-        options[key_content_view_id] ||= content_view_id(
-          scoped_options("content_view", options)
-        )
-      rescue HammerCLIForeman::MissingSeachOptions # rubocop:disable HandleExceptions
-        # Intentionally suppressing the exception,
-        # content_view is not always required.
-      end
+      options[key_content_view_id] ||= search_and_rescue(
+       :content_view_id, "content_view", options)
+      # Intentionally suppressing the exception,
```
bbuckingham: since the exception handling is now in search_and_rescue, we can probably move this comment there

bbuckingham:
@zjhuntin, All of the above scenarios that I mentioned are now working properly (without error).   I was also able to verify promotion of content view versions as raised by the original issue.  Nice!

lib/hammer_cli_katello/id_resolver.rb
```diff
@@ -93,23 +93,13 @@ def content_view_version_id(options)
...
 
-      begin
-        options[key_content_view_id] ||= content_view_id(
-          scoped_options("content_view", options)
-        )
-      rescue HammerCLIForeman::MissingSeachOptions # rubocop:disable HandleExceptions
-        # Intentionally suppressing the exception,
-        # content_view is not always required.
-      end
+      options[key_content_view_id] ||= search_and_rescue(
```
bbuckingham: this could be done on 1 line and rubocop would still be happy :)

zjhuntin:
This PR fixes two separate issues with how the lifecycle-environment IDs are resolved:

1. When trying to promote a content-view version with a named lifecycle-environment
2. When building  options for any command that can use lifecycle-environment as an option. 

To test promote a content-view version with a named lifecycle-environment, and also try to list the content-views for that lifecycle environment or just for an organization.

mccun934:
@bbuckingham ping on a re-review

bbuckingham:
@zjhuntin, ACK.  Thanks!

PR #379, opened 2016-04-12
fixes #14596 - content-view puppet-module add: disable id resolver

bbuckingham:
The id resolver was getting executed when executing the above
command with the --name parameter.  For this command, we do not
want that behavior.

### DIFF
--- a/lib/hammer_cli_katello/content_view_puppet_module.rb
+++ b/lib/hammer_cli_katello/content_view_puppet_module.rb
@@ -33,6 +33,14 @@ module HammerCLIKatello
       failure_message _("Could not add the puppet module")
 
       build_options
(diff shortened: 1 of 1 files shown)


komidore64:
i'm cool with this.

ACK

komidore64:
Come to think of it, I do have some questions.

This doesn't cause hammer to error or do anything weird? Is there any kind of output in the log that might look weird to a user?

bbuckingham:
@komidore64, I am not seeing any errors in ~/.hammer/log/hammer.log or to standard out when running through the various scenarios.

komidore64:
awesome

komidore64:
ACK

PR #382, opened 2016-04-15
Fixes #14611 - Add content-view purge command to delete old versions

rclsilver:


lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    def execute
+      if option_count < 0
+        print_message _("Invalid value for --count parameter: value must be greater than 0.")
+        HammerCLI::EX_USAGE
+      else
+        view = resource_content_views.call(:show, :id => option_id)
+
+        # Sort versions by the number (1.0, 2.0, ...)
+        all_versions = view['versions'].sort! do |a, b|
```
adamruzicka: Methods ending with exclamation mark usually modify the original object in place. In this case the `views['versions']` would be sorted too. It is not a problem here, since the `view['versions']` is not used anywhere else, but it could be overlooked easily

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -0,0 +1,60 @@
...
+          a["version"].to_f <=> b["version"].to_f
+        end
+
+        # Keep only versions which are not used
+        old_versions = all_versions.select! do |v|
```
adamruzicka: The same case as with the `sort!`

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -0,0 +1,60 @@
...
+          v["environment_ids"].empty?
+        end
+
+        # Check if there is something to do
+        if option_count >= old_versions.size
+          print_message _("No version to delete.")
+          HammerCLI::EX_OK
```
adamruzicka: This can be safely removed, since there is another expression after the whole if-else-end block

akofink:
NACK - no tests! :see_no_evil:

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -0,0 +1,59 @@
+module HammerCLIKatello
+
+  class ContentViewPurge < HammerCLIKatello::Command
```
akofink: Perhaps this file should be put in `hammer_cli_katello/content_view.rb` with the rest of the content_view hammer commands. Only subcommands get their own file usually (i.e. `hammer_cli_katello/content_view_puppet_module.rb`)

akofink:
Ping @Katello/cli If we want to support this, which I think makes sense, we should consider supporting it in the Katello API and remove the logic from hammer-cli-katello.

ehelms:
I'm OK with this and then moving it to the API at some point. I'm also good with going the pure API route.  @akofink do you have a strong preference?

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -0,0 +1,59 @@
...
+    def execute
+      if option_count < 0
+        print_message _("Invalid value for --count parameter: value must be greater than 0.")
+        HammerCLI::EX_USAGE
+      else
+        view = resource_content_views.call(:show, :id => option_id)
+
+        # Sort versions by the number (1.0, 2.0, ...)
+        all_versions = view['versions'].sort do |a, b|
+          a["version"].to_f <=> b["version"].to_f
```
akofink: I'm not sure this would work. Suppose you have two versions 14.9 and 14.10; I think this comparison would tell us that 14.9 > 14.10 because 14.9 > 14.1 (to_f).

You could sort by the two columns `major` and `minor`:

```
all_versions = view['versions'].sort_by do |version|
  [version[:major], version[:minor]]
end
```

I believe this would work.

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -0,0 +1,59 @@
...
+        old_versions = all_versions.select do |v|
+          v["environment_ids"].empty?
+        end
+
+        # Check if there is something to do
+        if option_count >= old_versions.size
+          print_message _("No version to delete.")
+        else
+          # Keep only versions to delete in the array
+          oldest_version = old_versions.slice(0, old_versions.size - option_count)
```
akofink: `oldest_version` seems like a single version, but it is in fact an array of versions. Perhaps `versions_to_purge` or something similar would be better?

akofink:
@ehelms I think we could implement an API endpoint to list old versions that don't have a lifecycle environment associated with them. The hammer command would issue that call, confirm with the user by showing the list of versions to delete (we could add a -y option to skip asking, similar to Yum/DNF), then issue a bulk delete call with those IDs.

akofink:
@rclsilver If you pull the latest upstream, you will be able to see test coverage on this PR.

rclsilver:
@akofink Thanks, but do I have something to do?

akofink:
@rclsilver Yes; please check out the line comments.

ehelms:
@rclsilver planning to revisit this PR? If not, @akofink would what you like to do with this?

akofink:
I think it's a useful command to have, and I don't mind implementing it in hammer rather than the API. If @rclsilver doesn't want to continue with it, we can triage it as an RFE and use this PR as a starting point for whoever gets it.

akofink:
Closing this PR and moving it to https://github.com/Katello/hammer-cli-katello/pull/491

PR #384, opened 2016-04-22
Fixes #14708 - Content-view remove properly works.

zjhuntin:
This fixes content-view remove, to test create a content-view with and promote a life cycle to a version you'd like to test. Then try removing it with hammer. Right now it cannot resolve the 'names' of content-view-versions. However this is being solved by #14604 by @akofink. The name is going to be  to be [name of content-view] [major-version].[minor-version].

### DIFF
--- a/lib/hammer_cli_katello/id_resolver.rb
+++ b/lib/hammer_cli_katello/id_resolver.rb
@@ -52,7 +52,7 @@ module HammerCLIKatello
     end
 
     def system_id(options)
-      options[HammerCLI.option_accessor_name("id")] || find_resource(:systems, options)['uuid']
+      options[HammerCLI.option_accessor_name("id")] || find_resource(:systems, options)["uuid"]
     end
 
     def environment_id(options)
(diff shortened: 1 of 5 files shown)


lib/hammer_cli_katello/id_resolver.rb
```diff
@@ -64,12 +64,18 @@ def environment_ids(options)
...
       end
 
-      key_names = HammerCLI.option_accessor_name 'lifecycle_environment_names'
-      key_organization_id = HammerCLI.option_accessor_name 'organization_id'
-      options[key_organization_id] ||= organization_id(scoped_options 'organization', options)
+      if options["option_lifecycle_environment_names"]
+        key_names = HammerCLI.option_accessor_name "lifecycle_environment_names"
+      elsif options["option_environment_names"]
+        key_names = HammerCLI.option_accessor_name "environment_names"
+      elsif options["option_names"]
```
mbacovsky: I'm a bit concerned about use of the unscoped 'names' here. If there is a command `some-resource --names x,y` that has also optional param `--lifecycle-environment-names` it would lead the command to resolve --names as environment_ids. I'm not sure if such situation can happen though.

PR #385, opened 2016-04-26
Fixes #14821: remove description from content host.

waldenraines:
Viewing/updating description on content host is no longer supported
as viewing/updating the comment field on host replaces this field.
This commit removes the description field output from hammer for
content hosts.

http://projects.theforeman.org/issues/14821


waldenraines:
Blocked by https://github.com/Katello/katello/pull/5998

jlsherrill:
ACK

PR #388, opened 2016-04-27
Fixes #14860 - Eratta apply has a success message that follows async command format.

zjhuntin:
To test just try to apply an errata to a content-host with the --async flag. It should response with the Errata is being applied with task [task-id].


jlsherrill:
ACK

PR #389, opened 2016-04-28
fixes #14871 - capsule content synchronize - fix error on using name

bbuckingham:
Before:
```
hammer> capsule content synchronize --name katello.example.com
Could not synchronize capsule content:
    Error: undefined method `[]' for nil:NilClass
```
After:
```
hammer> capsule content synchronize --name katello.example.com
Could not synchronize capsule content:
    This request may only be performed on a Capsule that has the Pulp Node feature.
```

### DIFF
--- a/lib/hammer_cli_katello/capsule.rb
+++ b/lib/hammer_cli_katello/capsule.rb
@@ -70,7 +70,8 @@ module HammerCLIKatello
         build_options
       end
 
-      class SyncCommand < HammerCLIForemanTasks::AsyncCommand
+      class SyncCommand < HammerCLIKatello::SingleResourceCommand
+        include HammerCLIForemanTasks::Async
         include LifecycleEnvironmentNameResolvable
         resource :capsule_content, :sync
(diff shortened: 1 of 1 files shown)


johnpmitsch:
It works as you described, but why is there still a warning about pulp nodes since they have been removed? Does our API send this?

johnpmitsch:
@bbuckingham looks like we send that [here](https://github.com/Katello/katello/blob/1a0abd7eec6a9a2d31ebd2b272800ef5111e199e/app/controllers/katello/api/v2/capsule_content_controller.rb#L88-L93) I'll have to look into why that is happening. As far as this PR, it fixes the issue in hammer so ACK

johnpmitsch:
@bbuckingham sorry, ignore my previous messages (this is what happens when I review PRs before lunch) I was trying to sync the main katello server, not a capsule. A capsule syncs fine.

ACK (again)

bbuckingham:
@johnpmitsch, Thanks!

PR #392, opened 2016-05-09
Fixes #14970 - LifecycleEnvironment will be called for once.

zjhuntin:
To test just try calling for a list of content-views with a --lifecycle-environment [name] as an option with the -vd flags. You should only see one api call to environments be made.


akofink:
ACK - gotta get that extra 0.001% coverage! :cake:

jlsherrill:
Merging based on ACK

PR #393, opened 2016-05-10
Fixes #14994 - Add simplecov with 46% minimum

akofink:
Request for comments.

Simplecov is [even more popular](https://www.ruby-toolbox.com/categories/code_metrics) in the Ruby community than Rubocop in the category of code metrics.

When you run your tests,  you will see the coverage summary print on one line of text. For more details, you can view a file-by-file summary that is generated in HTML at coverage/index.html. I've used simplecov in the past and been very happy with it.

The current coverage is 52%. I encourage people to raise the minimum defined in the test_helper as the coverage increases.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,8 +10,8 @@ group :test do
   gem 'thor'
   gem 'minitest', '4.7.4'
   gem 'minitest-spec-context'
-  gem 'simplecov'
   gem 'mocha'
+  gem 'coveralls', require: false
 end
 
 # load local gemfile
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -1,3 +1,16 @@
+if RUBY_VERSION > "2.2"
+  # Coverage - Keep these two lines at the top of this file
+  require 'simplecov'
+  require 'coveralls'
+  SimpleCov.formatters = [SimpleCov::Formatter::HTMLFormatter, Coveralls::SimpleCov::Formatter]
+  SimpleCov.start do
+    minimum_coverage 46
+    refuse_coverage_drop
+    track_files "lib/**/*.rb"
+    add_filter '/test/'
+  end
+end
+
 require File.join(File.dirname(__FILE__), './task_helper.rb')
 require 'minitest/autorun'
 require 'minitest/spec'
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# hammer-cli-katello ![Travis # Status](https://travis-ci.org/Katello/hammer-cli-katello.svg?branch=master)
+# hammer-cli-katello ![Travis # Status](https://travis-ci.org/Katello/hammer-cli-katello.svg?branch=master) [![Coverage Status](https://coveralls.io/repos/github/Katello/hammer-cli-katello/badge.svg?branch=master)](https://coveralls.io/github/Katello/hammer-cli-katello?branch=master)
 
 Next-gen CLI tool for [Katello](http://katello.org) (Katello-specific commands)
 


ehelms:
Is 91% the current coverage?

ehelms:
Does coverage generation add time to running the test suite? If yes, is it a trivial amount of time?

ehelms:
Looking at the coverage report, this is only seeming to report on "tested" files and not on the codebase as a whole which to me gives a false pretense of our code coverage.

akofink:
For this project, there is no noticeable execution time difference, and I've never had a problem with coverage increasing execution time in the past for large projects with slow tests.

Using simplecov's `track_files` option, now coverage is about 53%, which sounds much more reasonable.

ehelms:
@daviddavis thoughts?

@akofink do you have a proposal on how we handle bumping the coverage metric? Is our goal to just never drop below the current value and try to remember to increase it over time? Is there a way we can more systemically bump it?

daviddavis:
Usually for tools like coveralls (e.g. https://github.com/fusor/fusor/pull/825), the CI environment will fail the PR if the code coverage drops. Any chance we can do that here?

ehelms:
I think this will do it is what @akofink is saying -- running 'rake' will cause the tests to fail if the coverage drops. We could test it by setting it to 60 and re-running.

akofink:
Yes, it does fail. I already tried it with 53% (edge case).

daviddavis:
I like checking for a decrease in coverage better than setting a minimum. If increase the code coverage and don't realize it or forget to raise the minimum, then someone is free to add more code without tests. That seems bad to me.

akofink:
If you add more code without tests, the coverage will drop. There is also an [option to disallow the coverage to drop at all](https://github.com/colszowka/simplecov#refuse-dropping-coverage), but it only works if you commit the coverage/.last_run.json file, which would need to be committed by travis each time. I don't see a good solution for that feature.

daviddavis:
@akofink did you look at the PR I linked? Coveralls will track coverage over time and [it uses simplecov](https://github.com/lemurheavy/coveralls-ruby/blob/master/Gemfile#L8).

ehelms:
So this is sort of a limitation of using Travis given that it has no memory between runs. Also, developers running locally would have a hard time gauging if they were meeting coverage constraints each time.

daviddavis:
> So this is sort of a limitation of using Travis given that it has no memory between runs.

Coveralls takes care of this and shows rather useful information about coverage over time and by file.

> Also, developers running locally would have a hard time gauging if they were meeting coverage constraints each time.

This is a problem but it's easy to open a PR and have coveralls run it. I think you can also set it up to test your local github fork.

ehelms:
If coveralls is like Travis in that it's free to use without limits, I'm good with that option. This may result in people having to force push and wait for reports more often but I guess we can monitor that. For other repositories that use Jenkins, using simplecov would work.

daviddavis:
I opened [a PR to demonstrate coveralls](https://github.com/Katello/hammer-cli-katello/pull/394). I think the message is messed up because it's trying to check for a decrease in coverage but it hasn't been run against master before.

Gemfile
```diff
@@ -12,6 +12,8 @@ group :test do
   gem 'minitest-spec-context'
   gem 'simplecov'
   gem 'mocha'
+  gem 'simplecov', require: false
```
daviddavis: Might want to remove line 13.

test/test_helper.rb
```diff
@@ -1,3 +1,13 @@
+# Coverage - Keep these two lines at the top of this file
+require 'simplecov'
+require 'coveralls'
+SimpleCov.formatters = [SimpleCov::Formatter::HTMLFormatter, Coveralls::SimpleCov::Formatter]
+SimpleCov.start do
+  minimum_coverage 52
+  refuse_coverage_drop
+  track_files "{lib,test}/**/*.rb"
```
daviddavis: It looks like simplecov is checking coverage for files in the test directory.

akofink: Well, it is right now. I guess that's checking that we're exercising all of our test code. I'm willing to remove the test directory from coverage if you think that would be better.

daviddavis: Are you sure? I am thinking this causes simplecov to check that we have tests for our tests.

akofink: Ok, I agree; I've removed it now.

daviddavis:
This looks good to me. Let's merge this on Friday or Monday to give others (ie @Katello/cli) a chance to review it.

daviddavis:
ACK. Thanks @akofink.

PR #394, opened 2016-05-11
Fixes #14994 - Adding coveralls to check test coverage

daviddavis:


test/test_helper.rb
```diff
@@ -4,6 +4,9 @@
 require 'mocha/setup'
 require 'hammer_cli'
 require 'hammer_cli_foreman/commands'
+require 'coveralls'
+
+Coveralls.wear! if RUBY_VERSION > 2.2
```
daviddavis: We run travis against multiple ruby versions but we only want coveralls to run once so here I am just having it run against 2.2.x ruby versions (or greater) which also matches katello-deploy.

test/test_helper.rb
```diff
@@ -4,6 +4,9 @@
 require 'mocha/setup'
 require 'hammer_cli'
 require 'hammer_cli_foreman/commands'
+require 'coveralls'
+
+Coveralls.wear! if RUBY_VERSION > "2.2"
```
daviddavis: We run travis against multiple ruby versions but we only want coveralls to run once so here I am just having it run against 2.2.x ruby versions (or greater) which also matches katello-deploy.

akofink:
You should be able to use the simplecov Coveralls formatter.

```
require 'simplecov'
require 'coveralls'
SimpleCov.formatters = [SimpleCov::Formatter::HTMLFormatter, Coveralls::SimpleCov::Formatter]
SimpleCov.start do
  minimum_coverage 52
  refuse_coverage_drop
  track_files "{lib,test}/**/*.rb"
end
```

PR #395, opened 2016-05-12
Fixes #14853 - Support organization options in host-collections

akofink:
As usual, we require organization only when a host collection name is specified
rather than an ID. This is not true of `host-collection list` command, as it
will list all host collections across all organizations if no organization
options are given.

In addition, the host-collection copy interface is updated to be consistent with
the rest of hammer, allowing searching for a host collection by name if
the organization is also specified. The new host collection name is now specified
with --new-name rather than --name.

Note that I added test files for the broken content-host subcommand and created
a BZ for fixing the broken commands [#15289](http://projects.theforeman.org/issues/15289).

There is also another broken command `host-collection hosts`. The bug report to fix that
is [#15290](http://projects.theforeman.org/issues/15290).

The host-collection add-host/remove-host commands are broken as well, always
reporting success even when the hosts in question were not found: [#15291](http://projects.theforeman.org/issues/15291)
addresses this bug.

### DIFF
--- a/lib/hammer_cli_katello/host_collection.rb
+++ b/lib/hammer_cli_katello/host_collection.rb
@@ -47,7 +47,7 @@ module HammerCLIKatello
         field :total_hosts, _("Total Hosts")
       end
 
-      build_options
+      build_options { |o| o.expand(:all).including(:organizations) }
     end
 
     class HostsCommand < HammerCLIKatello::ListCommand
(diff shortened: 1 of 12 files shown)


akofink:
[test] for coverage

akofink:
@Katello/cli This one's been here a while and needs review.

test/functional/host_collection/content_base_test.rb
```diff
@@ -0,0 +1,7 @@
+require_relative '../test_helper'
+require 'hammer_cli_katello/host_collection'
+
+module HammerCLIKatello
+  describe HostCollection::ContentBaseCommand do
+  end
+end
```
komidore64: Is this test file necessary?

test/functional/host_collection/install_content_base_test.rb
```diff
@@ -0,0 +1,7 @@
+require_relative '../test_helper'
+require 'hammer_cli_katello/host_collection'
+
+module HammerCLIKatello
+  describe HostCollection::InstallContentBaseCommand do
+  end
+end
```
komidore64: Is this empty test file needed?

komidore64:
What is the reason for included test files that don't contain any tests?

akofink:
@komidore64 Those commands are completely broken, and I've created BZs for them, noted in the PR description. Creating a setup for those tests is just me being thorough, since those commands are a part of the host collections hammer commands.

zjhuntin:
These changes LGTM, the boat load of code coverage is great!

beav:
the change itself is LGTM, but please use redmine issues instead of BZs.

Also, I am +1 for not including the stubbed tests, but I will go along with whatever you and @komidore64 decide.

akofink:
@beav @komidore64 I've deleted the empty tests, since two people agreed to not include them.

beav:
ACK from me

komidore64:
ACK

swag dog approves

PR #396, opened 2016-05-12
Fixes #15029 - Upgrade to rubocop 0.39

daviddavis:


komidore64:
I don't like how some stuff gets indented way over to the right in multiline arrays and stuff, but I can live with it if that's what everybody else is cool with.

APT
thor likes it

komidore64:
:boom: :boom: :boom:

PR #397, opened 2016-05-12
Fixes #15032 - Fixing repository list test

daviddavis:
The repository list command doesn't have a `--lifecycle-environment` option. It's because the hash is getting modified somehow when running tests but not the hammer command. Perhaps it's a different load order? Regardless, freezing the hash fixes it.


zjhuntin:
ACK, looks good to me thanks! @daviddavis

PR #398, opened 2016-05-13
Fixes #15013 - Repository commands can accept organization flags.

zjhuntin:
The repository upload, info and sync command now can accepts organization as flags if the repository-id needs to be looked up.

### DIFF
--- a/test/functional/content_view/version/incremental_update_test.rb
+++ b/test/functional/content_view/version/incremental_update_test.rb
@@ -26,7 +26,8 @@ describe 'content-view version incremental-update' do
 
     expect_foreman_task('3')
 
-    run_cmd(@cmd + params)
+    result = run_cmd(@cmd + params)
+    assert(result.exit_code, 0)
   end
 
   it "performs incremental update with update all hosts" do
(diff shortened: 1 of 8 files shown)


beav:
ACK

test/functional/content_view/version/incremental_update_test.rb
```diff
@@ -26,7 +26,8 @@
 
     expect_foreman_task('3')
 
-    run_cmd(@cmd + params)
+    result = run_cmd(@cmd + params)
+    assert(result.exit_code, 0)
```
akofink: @zjhuntin I hate to comment on an already-merged PR, but you want to use `assert_equals` here. Look at the [docs for `assert`](http://ruby-doc.org/stdlib-2.0.0/libdoc/minitest/rdoc/MiniTest/Assertions.html#method-i-assert).

test/functional/repository/info_test.rb
```diff
@@ -0,0 +1,46 @@
...
+    params = ["--id=#{repo_id}"]
+
+    ex = api_expects(:repositories, :show, "Get info") do |par|
+      par["id"] == repo_id.to_s
+    end
+
+    ex.returns({})
+
+    result = run_cmd(@cmd + params)
+    assert(result.exit_code, 0)
```
akofink: The `assert` misuse is in this file, too.

PR #391, opened 2016-05-08
Fixes #14007 - Typo in MissingSeachOptions

ranjan:
It depends on [theforeman/hammer-cli-foreman#238](https://github.com/theforeman/hammer-cli-foreman/pull/238)

PR #399, opened 2016-05-16
Fixes #15067 - Use Hammer-CLI-Foreman current repo instead of version.

zjhuntin:
Updated to use gemspec, and fix MissingSeachOptions.

This issue is breaking bats, because it is using the  Hammer-CLI-Foreman's current repository, so we should also use the repository when testing.

### DIFF
--- a/lib/hammer_cli_katello/id_resolver.rb
+++ b/lib/hammer_cli_katello/id_resolver.rb
@@ -141,7 +141,7 @@ module HammerCLIKatello
 
     def search_and_rescue(search_function, resource, options)
       self.send(search_function, scoped_options(resource, options))
-    rescue HammerCLIForeman::MissingSeachOptions # rubocop:disable HandleExceptions
+    rescue HammerCLIForeman::MissingSearchOptions # rubocop:disable HandleExceptions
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/id_resolver.rb
```diff
@@ -141,7 +141,7 @@ def from_lifecycle_environment_id(options)
 
     def search_and_rescue(search_function, resource, options)
       self.send(search_function, scoped_options(resource, options))
-    rescue HammerCLIForeman::MissingSeachOptions # rubocop:disable HandleExceptions
+    rescue HammerCLIForeman::MissingSearchOptions # rubocop:disable HandleExceptions
```
jlsherrill: Did you mean to update this rather than just test and merge https://github.com/Katello/hammer-cli-katello/pull/391 after this gets merged?

zjhuntin: I found that if I didn't include this the tests failed.

jlsherrill: ahhhh works4me

PR #400, opened 2016-05-16
Fixes #14007 - Typo in MissingSeachOptions

zjhuntin:
Adds hammer-foreman repo.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ gemspec
 
 # for generating i18n files, gettext > 3.0 dropped ruby 1.8 support
 gem 'gettext', '>= 3.1.3', '< 4.0.0'
+gem 'hammer_cli_foreman', :git => 'https://github.com/theforeman/hammer-cli-foreman.git'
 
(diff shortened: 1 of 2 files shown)


beav:
ACK

komidore64:
Would you explain why we're locking down to ``hammer_cli_foreman`` master?

jlsherrill:
@komidore64 right now travis is running tests off the latest release of hammer_cli_foreman, while the nightly bats tests are using nightly builds.  So we are in a situation where hammer_cli_katello master isn't compatible with hammer_cli_foreman master, nor is hammer in katello nightly repos compatible with hammer in the foreman nightly repos.  

 I think in addition to this we should move to building hammer as a nightly release similar to foreman

komidore64:
who needs versions amirite?

zjhuntin:
Thanks @komidore64 and @ranjan!

beav:
I would be :+1: for combining some of the hammer repos, if they are tightly dependent on each other.

PR #401, opened 2016-05-20
Fixes #15313 - Extend hostgroups create/update commands

akofink:
- Add hostgroups create tests for hammer-cli-katello extensions
- Add hostgroups update tests for hammer-cli-katello extensions
- Extend hammer-cli-foreman to accept an organization_id when searching for a
  lifecycle_environment in the hostgroups subcommand
- Extend hammer-cli-foreman to accept an organization_id when searching for a
  content_view in the hostgroups subcommand

### DIFF
--- a/lib/hammer_cli_katello.rb
+++ b/lib/hammer_cli_katello.rb
@@ -126,4 +126,5 @@ module HammerCLIKatello
 
   # subcommands to hammer_cli_foreman commands
   require 'hammer_cli_katello/host'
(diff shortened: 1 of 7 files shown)


daviddavis:
Tested and it works. Code looks good. 👍 

ACK

daviddavis:
Actually should this be `Fixes #..` instead of `Refs #..`? It looks like it fixes the problem to me.

PR #402, opened 2016-05-23
Fixes #15142 - Replace improper assert() usage

akofink:
These methods come from MiniTest.

- `assert(x, msg)` asserts that x is true and prints msg when false.
- `assert_equal(x, y)` asserts that x and y are equal


jlsherrill:
ACK  thanks @akofink !

PR #403, opened 2016-05-25
Fixes #15189: Provide more available information for package list

ehelms:


akofink:
It would behoove us to write tests for the package command.

ehelms:
@akofink added some tests, one of which is failing for non-obvious reasons. Got some pointers as to why it fails? The command is a valid command.

test/functional/package/list_test.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  it "lists packages for a repository" do
+    params = [
+      "--product-id=#{product_id}",
+      "--repository-id=#{repo_id}"
+    ]
+
+    ex = api_expects(:packages, :index, 'repository packages list') do |par|
+      par['product_id'] == product_id &&
```
akofink: `product_id` isn't being passed to `Packages#index`; I don't see in the API where it is a param for [this endpoint](http://www.katello.org/docs/3.0/api/apidoc/packages.html).  Also, is there a reason you are validating the default page and per_page options? I believe we have [pagination tests](https://github.com/theforeman/hammer-cli-foreman/blob/master/test/functional/commands/list_test.rb#L4) that ensure those parameters are properly passed.

ehelms:
@akofink thanks for the help, updated

test/helpers/organization_helpers.rb
```diff
@@ -0,0 +1,15 @@
+module OrganizationHelpers
```
akofink: Why not `test/functional/helpers/organization_helpers.rb`? I think we're considering API expectations functional tests, and that file already exists. I don't see this helper being used in the unit tests.

ehelms:
@akofink moved to your idea

akofink:
@ehelms Sorry, that was the wrong path I gave. Look at the existing [organization_helpers.rb](https://github.com/Katello/hammer-cli-katello/blob/master/test/functional/organization/organization_helpers.rb#L2).

ehelms:
@akofink arg, no wonder I missed it, I would not think to look there. I am going to reconcile them all.

akofink:
@ehelms If you want to re-architect the helpers, be my guest. Just don't miss anything. ;)

ehelms:
@akofink moved everything to test/functional/helpers and updated all the paths

akofink:
Yes, I like that location. Thanks, @ehelms! APJ

akofink:
@ehelms I think pulling master will fix those paginate_by_default failures.

ehelms:
Updated this PR based on master and is all green now.

PR #404, opened 2016-05-26
Fixes #11535 - add facet info to host commands

jlsherrill:
this adds content and subscription facet info
to the host info and host list commands


test/functional/host/extensions/info_test.rb
```diff
@@ -0,0 +1,32 @@
...
+      par['id'] == '1'
+    end
+
+    json_file = File.join(File.dirname(__FILE__), 'data', 'host.json')
+    ex.returns(JSON.parse(File.read(json_file)))
+
+    param_ex = api_expects(:parameters, :index, 'Parameter index') do |par|
+      par[:host_id] == 1
+    end
+    param_ex.returns('results' => [])
```
akofink: There's an [`index_response` helper](https://github.com/theforeman/hammer-cli-foreman/blob/a7f86beb465737a4e8c4908a0fb9abfe824d0139/lib/hammer_cli_foreman/testing/api_expectations.rb#L53).

jlsherrill: sweet! updated

akofink:
Works as expected. Nice tests. ACK

jlsherrill:
Thanks for the review!

PR #408, opened 2016-06-13
Fixes #15395 - Repository name searches include organization-id.

zjhuntin:
To test, try creating a content-view filter using repository names.

### DIFF
--- a/lib/hammer_cli_katello/filter.rb
+++ b/lib/hammer_cli_katello/filter.rb
@@ -53,6 +53,15 @@ module HammerCLIKatello
       success_message _("Filter created")
       failure_message _("Could not create the filter")
 
+      validate_options do
+        organization_options = [:option_organization_id, :option_organization_name, \
+                                :option_organization_label]
+
+        if option(:option_repository_names).exist? || option(:option_content_view_name).exist?
+          any(*organization_options).required
+        end
+      end
+
       build_options
     end
 
--- a/lib/hammer_cli_katello/search_options_creators.rb
+++ b/lib/hammer_cli_katello/search_options_creators.rb
@@ -1,14 +1,26 @@
 module HammerCLIKatello
   module SearchOptionsCreators
+    def create_repository_search_options(options)
+      name = options[HammerCLI.option_accessor_name("name")]
+      organization_id = options[HammerCLI.option_accessor_name("organization_id")]
+
+      search_options = {}
+      search_options['name'] = name if name
+      search_options['organization_id'] = organization_id if organization_id
+      search_options
+    end
+
     def create_repositories_search_options(options)
       name = options[HammerCLI.option_accessor_name("name")]
       names = options[HammerCLI.option_accessor_name("names")]
       product_id = options[HammerCLI.option_accessor_name("product_id")]
+      organization_id = options[HammerCLI.option_accessor_name("organization_id")]
 
       search_options = {}
       search_options['name'] = name if name
       search_options['names'] = names if names
       search_options['product_id'] = product_id if product_id
+      search_options['organization_id'] = organization_id if organization_id
       search_options
     end
 
--- a/test/functional/activaton_key/list_test.rb
+++ b/test/functional/activaton_key/list_test.rb
@@ -1,8 +1,6 @@
 require File.join(File.dirname(__FILE__), '../test_helper')
 require File.join(File.dirname(__FILE__), '../lifecycle_environment/lifecycle_environment_helpers')
 
-require 'hammer_cli_katello/content_view_puppet_module'
-
 describe 'listing activation-keys' do
   include LifecycleEnvironmentHelpers
 
--- a/test/functional/content_view/create_test.rb
+++ b/test/functional/content_view/create_test.rb
@@ -1,4 +1,4 @@
-require File.join(File.dirname(__FILE__), '../test_helper')
+require_relative '../test_helper'
 
 # Workaround for issue #14289
 require 'hammer_cli_katello/content_view_puppet_module'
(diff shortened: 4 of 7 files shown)


lib/hammer_cli_katello/search_options_creators.rb
```diff
@@ -1,14 +1,26 @@
...
+    def create_repository_search_options(options)
+      name = options[HammerCLI.option_accessor_name("name")]
+      organization_id = options[HammerCLI.option_accessor_name("organization_id")]
+
+      search_options = {}
+      search_options['name'] = name if name
+      search_options['organization_id'] = organization_id if organization_id
+      search_options
+    end
+
```
akofink: You should add/modify the search_options_creators_test.rb

akofink:
Tests look great! Just one thing: please test that this validation is failing properly:

`Validation failed: cannot contain filters whose repositories do not belong to this content view`

akofink:
ACK - the test for that validation exists in the model.

lib/hammer_cli_katello/filter.rb
```diff
@@ -53,6 +53,13 @@ class CreateCommand < HammerCLIKatello::CreateCommand
       success_message _("Filter created")
       failure_message _("Could not create the filter")
 
+      validate_options do
+        organization_options = [:option_organization_id, :option_organization_name, \
+                                :option_organization_label]
+
+        any(*organization_options).required
+      end
```
komidore64: does the server not do this validation?

zjhuntin: Organization currently is not required in order to try and create a content-view filter, if you give it all IDs, it will currently require it if you try and give it the name of a content-view. 

However recently the API was changed to not require Organization-id when trying to look up the id of a repository. So if a person were to give a repository-name without an organization-id it would fail. 

I don't think that this causes any harm, but thinking about it now I could change it to only require organization_options if either content-view name or repository name is provided.

akofink: You could use `validate_options` - i.e. https://github.com/Katello/hammer-cli-katello/blob/364397877e80cf99d098748bb418476f1fea079a/lib/hammer_cli_katello/host_package_group.rb#L12

zjhuntin:
@komidore64 I modified the changes so that it will only require organization when content-view name or repository-name.

komidore64:
I'm still confused. I have the understanding that the server is responsible for telling you if you have all the proper params or not. Would you humor an old man and explain to me again why we need to validate this in hammer?

akofink:
@komidore64 I believe hammer has to validate those options because the API doesn't require them: https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/repositories_controller.rb#L42

komidore64:
``organization_id`` should be included when we're looking up a repository by name, I don't disagree with that.

But I don't think hammer should validate you include an organization reference when searching for a repository. That is the job of the server.

akofink:
@komidore64 There may be a good reason the API doesn't require organization_id, but for hammer, it needs to find only one repository, and name is only unique inside an organization.

It could be, however, that the API is not behaving as intended and should be changed (asking in #foreman-dev)

komidore64:
> There may be a good reason the API doesn't require organization_id, but for hammer, it needs to find only one repository, and name is only unique inside an organization.

I realize all of this. You would search for a repository by name, and the server would respond with more than one result, cause the ID lookup to respond with a message saying there was more than one result. "Okay, I need to try again and include the organization this time."

komidore64:
Is that not the desired flow?

akofink:
@komidore64 I see what you're saying, and you're right. I do think the error message that specifically says something about `--organization options` is more clear, but the user could also restrict the search (potentially) with other options that aren't organization and end up with a valid single result.

lib/hammer_cli_katello/filter.rb
```diff
@@ -53,6 +53,15 @@ class CreateCommand < HammerCLIKatello::CreateCommand
       success_message _("Filter created")
       failure_message _("Could not create the filter")
 
+      validate_options do
+        organization_options = [:option_organization_id, :option_organization_name, \
+                                :option_organization_label]
+
+        if option(:option_repository_names).exist? || option(:option_content_view_name).exist?
+          any(*organization_options).required
```
akofink: So @zjhuntin I think we agree to remove this validation of organization options, unless you have objections.

zjhuntin: I'm against it only because I think that the message that Hammer will give is much more clear. It lets the user know that they can specify organization-id, label or name.

I'm pretty sure the error that they get back from the API will not have as much information, and be more confusing. The message tailored here I believe gives a better use experience.

zjhuntin:
>  You would search for a repository by name, and the server would respond with more than one 
>  result, cause the ID lookup to respond with a message saying there was more than one result. "Okay, >  I need to try again and include the organization this time."

In this situation it would not. This command can take multiple repositories. When the request is made to the server with all of these names it doesn't generate the URL so that it will only search for these names. It actually ends up returning the names of every repository. This means that if another organization had a similarly named repository, it would be returned and could potentially be picked. It would still fail because the repository doesn't belong to that content-view, but would not give off the proper error message.

lib/hammer_cli_katello/filter.rb
komidore64: ok

komidore64:
i'm fine with this, but i think it should still be the server's responsibility to validate the params. if it's error messages aren't good enough, then hopefully in the future hammer will better catch thrown server errors and now how to display the proper information.

akofink:
So, @komidore64, ready to merge?

komidore64:
yeah

PR #410, opened 2016-06-15
Fixes #15429: use host API for host collection hosts.

waldenraines:
The host collection hosts command was pointing at an API that had been
removed.  This change brings the command in line with the API.

http://projects.theforeman.org/issues/15429

### DIFF
--- a/lib/hammer_cli_katello/host_collection.rb
+++ b/lib/hammer_cli_katello/host_collection.rb
@@ -51,15 +51,45 @@ module HammerCLIKatello
     end
 
     class HostsCommand < HammerCLIKatello::ListCommand
-      resource :host_collections, :hosts
+      resource :hosts, :index
       command_name "hosts"
+
+      option "--id", "HOST_COLLECTION_ID", _("Host Collection ID"),
+             :attribute_name => :option_host_collection_id
+      option "--name", "HOST_COLLECTION_NAME", _("Host Collection Name"),
+             :attribute_name => :option_host_collection_name
+
+      validate_options do
+        host_collection_options = [:option_host_collection_id, :option_host_collection_name]
+        any(*host_collection_options).required
+      end
+
+      def host_collection_options
+        {
+          "option_name" => option_host_collection_name,
+          "option_organization_name" => option_organization_name,
+          "option_organization_id" => option_organization_id,
+          "option_organization_label" => option_organization_label
+        }
+      end
+
+      def request_params
+        params = super
+        host_collection_id = option_host_collection_id
+        unless option_host_collection_name.nil?
+          host_collection_id = resolver.host_collection_id(
+            options.merge(resolver.scoped_options("host_collection", options)))
+        end
+        params['search'] = "host_collection_id=#{host_collection_id}"
+        params
+      end
 
       output do
         field :id, _("ID")
         field :name, _("Name")
       end
 
-      build_options
+      build_options { |o| o.expand(:all).including(:organizations) }
     end
 
     class CopyCommand < HammerCLIKatello::CreateCommand
--- a/test/functional/host_collection/hosts_test.rb
+++ b/test/functional/host_collection/hosts_test.rb
@@ -1,0 +1,40 @@
+require_relative '../test_helper'
+require 'hammer_cli_katello/host_collection'
+
+module HammerCLIKatello
+  describe HostCollection::HostsCommand do
+    it 'allows host collection id' do
+      api_expects(:hosts, :index) do |par|
+        par['search'] == "host_collection_id=1"
+      end
+
+      run_cmd(%w(host-collection hosts --id 1))
+    end
+
+    it 'allows host collection name' do
+      ex = api_expects(:host_collections, :index)
+      ex.returns(index_response([{'id' => 2}]))
+
+      api_expects(:hosts, :index) do |par|
+        par['search'] == "host_collection_id=2"
+      end
+
+      run_cmd(%w(host-collection hosts --name collection --organization-id 1))
+    end
+
+    it 'requires organization with host collection name' do
+      result = run_cmd(%w(host-collection hosts --name collection))
+      expected_error = "Error: Could not find organization, please set one of options " \
+                       "--organization, --organization-label, --organization-id."
+      assert_equal(HammerCLI::EX_SOFTWARE, result.exit_code)
+      assert_equal(expected_error, result.err[/#{expected_error}/])
+    end
+
+    it 'requires host collection name or id' do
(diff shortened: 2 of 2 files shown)


akofink:
I think you should support the `--id` and `--name` options here for the host collection, just for consistency. 

It may be good to have your tests reflect that these are valid options. Just make sure that you're not putting tests in hammer-cli-katello that should be or already are in hammer-cli-foreman. Furthermore, it would be good to have a negative-case test, in which you assert failure when all required options are not present.

waldenraines:
@akofink updated.

lib/hammer_cli_katello/host_collection.rb
```diff
@@ -51,15 +51,46 @@ class InfoCommand < HammerCLIKatello::InfoCommand
...
+        end
+      end
+
+      def host_collection_options
+        {
+            "option_name" => option_name,
+            "option_organization_name" => option_organization_name,
+            "option_organization_id" => option_organization_id,
+            "option_organization_label" => option_organization_label
+        }
```
waldenraines: Please let me know if there is a better way to do this.  This seems strange to recreate what I would think should already be there in some existing options object that I am unaware of.

lib/hammer_cli_katello/host_collection.rb
```diff
@@ -51,15 +51,47 @@ class InfoCommand < HammerCLIKatello::InfoCommand
...
+        {
+          "option_name" => option_name,
+          "option_organization_name" => option_organization_name,
+          "option_organization_id" => option_organization_id,
+          "option_organization_label" => option_organization_label
+        }
+      end
+
+      def request_params
+        params = {}
```
akofink: You should probably set `params = super` here.

lib/hammer_cli_katello/host_collection.rb
```diff
@@ -51,15 +51,47 @@ class InfoCommand < HammerCLIKatello::InfoCommand
...
+      option "--name", "HOST_COLLECTION_NAME", _("Host Collection Name")
+
+      validate_options do
+        host_collection_options = [:option_id, :option_name]
+        organization_options = [:option_organization_id, :option_organization_name,
+                                :option_organization_label]
+
+        any(*host_collection_options).required
+
+        if option(:option_name).exist?
```
akofink: I think you can rely on the `host_collections#index` search to ensure organization is required when the host collection ID has yet to be resolved from the host name.

akofink:
Functionality works, and the tests seem to test the relevant functionality. Just a few small comments.

lib/hammer_cli_katello/host_collection.rb
waldenraines: I changed that in order to get past hammer looking up the host with the `--name` parameter rather than the host collection.  Using `params = super` causes --name to be used to look up the host API instead of the host collections API in other words.  Do you have any other ideas to get around that while using `params = super`?

akofink: I think you can set the attribute_name when defining the option, like [this](https://github.com/Katello/hammer-cli-katello/blob/364397877e80cf99d098748bb418476f1fea079a/lib/hammer_cli_katello/content_view_version.rb#L76). This allows `--name` to be given to hammer as `option_host_collection_name`

lib/hammer_cli_katello/host_collection.rb
```diff
@@ -51,15 +51,47 @@ class InfoCommand < HammerCLIKatello::InfoCommand
...
+          any(*organization_options).required
+        end
+      end
+
+      def host_collection_options
+        {
+          "option_name" => option_name,
+          "option_organization_name" => option_organization_name,
+          "option_organization_id" => option_organization_id,
+          "option_organization_label" => option_organization_label
```
waldenraines: Is there a better, less repetitive, way to do this?

lib/hammer_cli_katello/host_collection.rb
waldenraines: 👍

lib/hammer_cli_katello/host_collection.rb
waldenraines: For some reason if I do that then the exception ends up being `EX_SOFTWARE` instead of `EX_USAGE` in the test.

waldenraines: Which seems wrong but is [apparently expected](https://github.com/waldenraines/hammer-cli-katello/blob/c4f0d8c8897ef69f91fb464fbdc27af4174a30b1/test/functional/host_collection/info_test.rb#L14-14).

waldenraines:
@akofink updated.

test/functional/host_collection/host_test.rb
```diff
@@ -0,0 +1,40 @@
+require_relative '../test_helper'
```
akofink: I think this file should technically be `hosts_test.rb` since the subcommand is `hosts`.

lib/hammer_cli_katello/host_collection.rb
```diff
@@ -51,15 +51,40 @@ class InfoCommand < HammerCLIKatello::InfoCommand
...
+      end
+
+      def request_params
+        params = super
+        collection_id = resolver.host_collection_id(host_collection_options) if option_host_collection_name
+        collection_id = option_host_collection_id if option_host_collection_id
+        params['search'] = "host_collection_id=#{collection_id}"
+        params
+      end
+
```
akofink: I think this works, but the test API data doesn't seem to be properly mirroring the actual API. You don't need to limit the options passed to `host_collection_id()`:

```ruby
   class HostsCommand < HammerCLIKatello::ListCommand
      resource :hosts, :index
      command_name "hosts"

      option "--id", "HOST_COLLECTION_ID", _("Host Collection ID"),
        attribute_name: :option_host_collection_id
      option "--name", "HOST_COLLECTION_NAME", _("Host Collection Name"),
        attribute_name: :option_host_collection_name

      def request_params
        params = super
        option_host_collection_id ||= resolver.host_collection_id(options.merge(resolver
          .scoped_options("host_collection", options))) if option_host_collection_name
        params['search'] = "host_collection_id=#{option_host_collection_id}"
        params
      end

      output do
        field :id, _("ID")
        field :name, _("Name")
      end

      build_options { |o| o.expand(:all).including(:organizations) }
    end
```

waldenraines:
@akofink updated again.

akofink:
Looks good! Thanks @waldenraines! ACK

waldenraines:
@akofink please merge, I don't have the permissions.

waldenraines:
Or @komidore64, etc.

PR #411, opened 2016-06-16
Fixes #15436 - Update test API and tests

akofink:
This needs to be modified once the test API was refreshed due to
https://git.io/voEtA (katello/katello f4ae98f), which changes to API to
not require organization options when searching for a repository


ehelms:
ACK - thanks @akofink !

PR #412, opened 2016-06-20
Fixes #15458: allow org id for activation key host collections.

waldenraines:
The add/remove host collection commands were not allowing
--organization-id.  This commit ensures that all organization related
parameters are accepted.

http://projects.theforeman.org/issues/15458


waldenraines:
> Coverage decreased (-0.07%) to 60.685%

On an untested file 👎

cfouant:
Works for me, ACK! Nice and elegant, good job Walden!

PR #413, opened 2016-06-22
Fixes #14604 - Add ID resolution to CV add-version

akofink:
- ContentViewVersions can now be added to a ContentView via ContentView
ID and ContentViewVersion version.

Like this:

```sh
hammer content-view add-version --id 2 --content-view-version 1.0 --content-view-version-content-view-id 3
hammer content-view remove-version --id 2 --content-view-version 1.0 --content-view-version-content-view-id 3
```


daviddavis:
ACK

PR #414, opened 2016-06-28
Fixes #15537 - Allow package names as an array

akofink:
This change allows users to specify package or package group names in a
comma-separated list when creating a content view filter rule rather
than one package at a time.

Prerequisite: https://github.com/Katello/katello/pull/6157


bbuckingham:
ACK once https://github.com/Katello/katello/pull/6157 is merged.

PR #415, opened 2016-06-29
Fixes #15540 - Allow repo name in remove-content

akofink:
- The `--name` option now resolves the repository id when removing content
from a custom repository.
- The ID resolver for repository_id no longer requires product options.


zjhuntin:
For doing what this is doing ACK.  Though as we talked about there should be some safeguard/option to specify by product and/or organization.

akofink:
@zjhuntin This is the issue to add organization/product options: http://projects.theforeman.org/issues/15555

PR #416, opened 2016-07-06
Fixes #15589 - Remove content-host update command

akofink:
The new command is `hammer host update ...` which exists in hammer-cli-foreman/Foreman.

### DIFF
--- a/lib/hammer_cli_katello/content_host.rb
+++ b/lib/hammer_cli_katello/content_host.rb
@@ -51,23 +51,6 @@ module HammerCLIKatello
       build_options
     end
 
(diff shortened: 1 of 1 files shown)


bbuckingham:
This would be handled separate from this PR; however, are we tracking an issue to remove the update action from the Katello API ?   (ref: https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/systems_controller.rb#L71)

bbuckingham:
APT

akofink:
Tests will be dependent on merging https://github.com/Katello/hammer-cli-katello/pull/417

akofink:
[test]

PR #418, opened 2016-07-07
Fixes #15604 - Handle nil repo list in repo_ids

akofink:
This change allows the `repository_ids` ID resolver to handle a nil
collection of repository names.


bbuckingham:
APT

PR #422, opened 2016-07-15
Fixes #15701 - Resolve puppet environment

akofink:
Resolve puppet environment name from api/environments (puppet
environments) rather than katello/api/environments (lifecycle
environments)

@ehelms Not quite the solution we discussed last night because the superclass environment_id method doesn't actually exist, so an alias doesn't work - The IdResolver in both katello and foreman will resolve methods that aren't defined to use the get_resource() method. I'm using get_id, which uses get_resource, so the puppet environment ID should be getting resolved just as it was before we changed the resolver to HammerCLIKatello::IdResolver.


akofink:
To test, try the following:

```
hammer -vd hostgroup create --environment puppetenv --name hostgroup1
```

Ensure the environment is resolving using api/environments and not katello/api/environments

ehelms:
ACK

lzap:
The patch does not work: https://bugzilla.redhat.com/show_bug.cgi?id=1354462#c5

When testing this, make sure you have more than one puppet environment and architecture.

lzap:
The other patch fixed it, thanks!

PR #423, opened 2016-07-18
Refs #15701: Resolve environments without Katello API using search

ehelms:


ehelms:
To ensure that when doing something like hostgroup create, the Puppet environments API is searched instead of using a parameter. In other words, using debug mode, we want to see:

```
[DEBUG 2016-07-11T22:15:32 API] Params: {
    :search => "name = \"production\""
}   
[DEBUG 2016-07-11T22:15:32 API] Headers: {
    :params => {
        :search => "name = \"production\""
    }   
} 
```

akofink:
Not sure how we missed that one - looks good, @ehelms ACK

test/unit/search_options_creators_test.rb
```diff
@@ -106,6 +106,12 @@ def two
       end
     end
 
+    describe '#create_environments_search_options(options)' do
+      it 'does not use the katello api' do
+        search_options_creators.create_organizations_search_options(:environments)
```
akofink: You want this to be `create_environments_search_options`

ehelms:
@akofink added tests

PR #424, opened 2016-07-18
Refs #12556 - remove system subcommand

jlsherrill:


akofink:
What about https://github.com/Katello/hammer-cli-katello/blob/edafb34667e01c6ff9eac4489c2feb6fec081b35/lib/hammer_cli_katello/subscription.rb#L8

jlsherrill:
@akofink what about it?  :) 

That's not related to the removal of systems, unless i'm misunderstanding what you mean.

akofink:
@jlsherrill And this https://github.com/Katello/hammer-cli-katello/blob/e3a6acb8ca068288eee1cf8b2b89842d168d7d6e/lib/hammer_cli_katello/id_resolver.rb#L63

I don't think the `:systems` resource exists anymore, so any reference to it will cause NoMethodError on nil.

akofink:
@jlsherrill Is this the PR to fix any hammer issues caused by the host unification Katello PR?

akofink:
You should rebuild the apipie-cache and put it on this PR - some of the test failures will point out other uses of the systems resource.

jlsherrill:
@akofink the 'broken' command is tracked here: http://projects.theforeman.org/issues/14403

This is simply the PR to remove all uses of the systems controller  (/katello//api/v2/systems/)

jlsherrill:
Will do!

jlsherrill:
@akofink added!

akofink:
Looks good. ACK Thanks @jlsherrill!

PR #425, opened 2016-07-19
Refs #12556 - fix content host reference

jlsherrill:


akofink:
Disregarding the small coverage drop - ACK

PR #426, opened 2016-07-28
Fixes #15889 - Resolve IDs when there are names

akofink:
Resolve repository_ids only when there are repository_names

I expect upstream tests to fail, but the `id_resolver_test` should pass.


bbuckingham:
@akofink, good work.  ACK!

PR #427, opened 2016-07-29
Fixes #15733 - cleanup unused option organization-id from content view version list/info

ranjan:


akofink:
This seems to remove some important options like `--content-view` and others:

Before
```
Usage:
    hammer content-view version info [OPTIONS]

Options:
 --content-view CONTENT_VIEW_NAME                            Content view name to search by
 --content-view-id CONTENT_VIEW_ID                           Content view to search by
 --content-view-organization-id CONTENT_VIEW_ORGANIZATION_ID Organization ID
 --environment ENVIRONMENT_NAME                              Name to search by
 --environment-id ENVIRONMENT_ID                              
 --id ID                                                     Content view version identifier
 --organization ORGANIZATION_NAME                            Organization name to search by
 --organization-id ORGANIZATION_ID                           organization ID
 --organization-label ORGANIZATION_LABEL                     Organization label to search by
 --version VERSION                                           Content view version number
 -h, --help                                                  print help
```

After
```
Usage:
    hammer content-view version info [OPTIONS]

Options:
 --content-view-id CONTENT_VIEW_ID Content view to search by
 --id ID                           Content view version identifier
 --version VERSION                 Content view version number
 -h, --help                        print help
```

Furthermore, I'm not convinced we should "fix" this issue by removing organization options. If you have multiple content views with the same name in different organizations, you want to be able to filter by organization (i.e. `--content-view cv1 --organization 'Default Organization'`.

ranjan:
@akofink,  
I looked into content_view_**version** api's and it do not support options called organization so thought to remove this options for content-view version info . 

As per my understanding(Regarding info)
 1. Info command is designed to handle only one record(from api response) when we allow options other than id it raises error like *multiple record found*(incase content view version).
 2. Since id is mandatory options for info command others options will be silently ignored(because api will return record based on content-view-version-id)
Feel free to close this pull request if you feel organization still required for the content view version.

akofink:
@ranjan Other options besides what is in the API can be hammer options because we allow the IDs of most models to be resolved by (at least) a search on their name. For example, a content view can be found by name, but name is only unique to an organization_id, so we also allow organization_id. Instead of requiring our users to know the ID of their organization, we allow them to use a name such as 'Default Organization'. Hammer makes three requests in this case, one to organizations#index which returns the organization ID and then a request to content_views#index to get the content_view ID, then finally a request to content_view#show with the resolved organization_id and content_view id.

I do want to point out, however, that organization_id is not required when searching for a content_view by name, but it *may* be if there are identically named content views in multiple organizations because the request to resolve the content view expects only one unique record to be returned.

ranjan:
Thanks @akofink for explaining the issue. closing this now.

PR #428, opened 2016-07-29
Fixes #15830 - Handle non-required org options

akofink:
This is a far-reaching change, mostly due to the API update which removed required organization options on `content_views#index` and `lifecycle_environments#index`.

This is a large part of the current bats failure.

Please review and test carefully as this fixes many different commands, and it would be easy to overlook a command that doesn't have automated tests.


test/functional/hostgroup/create_test.rb
```diff
@@ -28,7 +28,7 @@ module HammerCLIForeman
         api_expects(:hostgroups, :create) do |p|
           p['hostgroup']['content_view_id'] == 1 && p['hostgroup']['name'] == 'hg1'
         end
-        run_cmd(%w(hostgroup create --name hg1 --content-view cv1 --content-view-organization-id 1))
+        run_cmd(%w(hostgroup create --name hg1 --content-view cv1 --query-organization-id 1))
```
akofink: Note this new command syntax for creating/updating hostgroups, as recommended by @bbuckingham

akofink:
I've added the majority of the bats tests in this PR as functional tests to prevent hammer-cli-katello PRs that break bats to be merged because nightly builds are important. :night_with_stars:

jlsherrill:
ACK tested with a variety of commands and seemed to work fine.   Code looks reasonable

PR #433, opened 2016-08-02
Refs #15585 - hosts#index doesn't call parameters

akofink:
`hammer host info` used to make a call to /hosts/:id/parameters, but
now it doesn't.

Regression caused by https://github.com/theforeman/hammer-cli-foreman/pull/249

Co-requisite: https://github.com/Katello/hammer-cli-katello/pull/434

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: ruby
 rvm:
-  - 2.0.0
   - 2.1
   - 2.2
 before_install:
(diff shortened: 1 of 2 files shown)


komidore64:
It seems odd that we're testing stuff that is in `hammer-cli-foreman`? Is that correct @akofink?

akofink:
The build is failing due to https://github.com/theforeman/hammer-cli/pull/198 :cry:

akofink:
@komidore64 We extend hosts, so we should test those extensions but not anything hammer-cli-foreman tests.

komidore64:
We *should*, but are we?

akofink:
@komidore64 I'm not writing a test here, just removing some tested logic that no longer applies. @jlsherrill wrote this test originally.

komidore64:
ack pending travis

cfouant:
APT

PR #434, opened 2016-08-02
Fixes #15949 - Drop Ruby 2.0.0

akofink:
co-requisite: https://github.com/Katello/hammer-cli-katello/pull/433


akofink:
This commit moved to https://github.com/Katello/hammer-cli-katello/pull/433

PR #435, opened 2016-08-04
Fixes #15987 - Removing fields for space and consistency

daviddavis:
We don't show these fields for other content types like RPMs so remove them to save space.


bbuckingham:
APT

akofink:
I'd approve changing the test_helper to remove https://github.com/Katello/hammer-cli-katello/blob/master/test/test_helper.rb#L9

PR #436, opened 2016-08-05
Fixes #16000 - Show correct state in listing product content for an activation key.

ranjan:


ranjan:
It depends on katello/katello#6232

lib/hammer_cli_katello/activation_key.rb
```diff
@@ -40,7 +40,8 @@ def request_params
         params = super
         if options.keys.any? { |o| o.match(/\Aoption_organization.*/) }
           params['organization_id'] = resolver.organization_id(
-            resolver.scoped_options('organization', all_options))
+            resolver.scoped_options('organization', all_options)
+          )
```
akofink: This seems like an unnecessary change.

akofink:
The change looks good, but please add a test for the product-content sub-command.

lib/hammer_cli_katello/activation_key.rb
ranjan: The above indentation change is suggested by rubocop.

akofink:
LGTM - ACK Thanks @ranjan!

ranjan:
Thanks @akofink, Just for information Is it ok to merge this commit without waiting for the PR Katello/katello#6232 to be merged?.

akofink:
@ranjan Good catch; That's my bad. The tests will still pass on hammer-cli-katello because they run against a cached API, but it will break master branch for hammer until Katello/katello#6232 is merged. As for downstream, the two changes will be cherry-picked in together because they're on the same issue.

For the future, it helps to put a [WIP] in the title of the hammer PR until the dependencies are met.

PR #439, opened 2016-08-09
Fixes #16034 - Org resolution in AK add-host-coll

akofink:
Ensure organization ID resolution behaves properly in `hammer
activation-key add/remove-host-collection`


bbuckingham:
ACK

PR #442, opened 2016-08-12
Fixes #16101 - Update Gemfile and README setup info

daviddavis:
Looking for feedback on this. If you follow the setup instructions in the README, you get a bundler error:

```
[!] There was an error parsing `Gemfile`: You cannot specify the same gem twice coming from different sources.
You specified that hammer_cli (>= 0) should come from https://github.com/theforeman/hammer-cli.git (at master) and source at `../hammer-cli`
. Bundler cannot continue.

 #  from /home/vagrant/Projects/hammer-cli-katello/Gemfile:25
 #  -------------------------------------------
 #    local_gemfile = File.join(File.dirname(__FILE__), file_name)
 >    self.instance_eval(Bundler.read_file(local_gemfile)) if File.exist?(local_gemfile)
 #  end
 #  -------------------------------------------
```

It's because we specify git repo locations for hammer_cli and hammer_cli_foreman. This PR removes the development group and documents how developers can use either local checkouts or git repos via Gemfile.local. This should also simplify the hammer-devel playbook (see https://github.com/Katello/forklift/pull/273).

PR #443, opened 2016-08-22
Refs #16101 - Remove gem git requirements

daviddavis:
The PR https://github.com/Katello/hammer-cli-katello/pull/442 didn't actually work because bundler won't let you specify different sources for a gem even if the gem requirements are in different groups. This PR simply writes out to Gemfile.local for travis using the git urls.


komidore64:
ack

PR #449, opened 2016-09-12
Fixes #16521 - Add support for glob paths

daviddavis:
Add support for globs when users upload files to repos.

### DIFF
--- a/lib/hammer_cli_katello/repository.rb
+++ b/lib/hammer_cli_katello/repository.rb
@@ -194,8 +194,10 @@ module HammerCLIKatello
 
           if File.directory?(fullpath)
             Dir["#{fullpath}/*"]
-          else
+          elsif File.exist?(fullpath)
             [fullpath]
+          else
+            Dir[fullpath]
           end
         end
       end
@@ -206,7 +208,14 @@ module HammerCLIKatello
 
       def execute
         @failure = false
-        option_content.each do |file_path|
+        files = option_content
+
+        if files.length.zero?
+          output.print_error _("Could not find any files matching PATH")
+          return HammerCLI::EX_NOINPUT
+        end
+
+        files.each do |file_path|
           File.open(file_path, 'rb') { |file| upload_file file }
         end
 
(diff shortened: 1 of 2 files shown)


akofink:
Is this intended behavior?

```sh
$ hh repository upload-content --id 4 --path '~/.{bar,bash,foo}rc'
Successfully uploaded file '.bashrc'.
```

Should we offer messages when files fail to upload (i.e. `~/.barrc, ~/.foorc`)

akofink:
Some more strange behavior for [nested] directories:

```
akofink@akofink-desktop: ~/dev/katello-devel/hammer-cli-katello temp/20160912120952
$ hh repository upload-content --id 4 --path '~/tmp/16521' 
Successfully uploaded file 'foo'.
Successfully uploaded file 'bar'.
Failed to upload file 'baz' to repository. Please check the file and try again.
FAIL: 65

akofink@akofink-desktop: ~/dev/katello-devel/hammer-cli-katello temp/20160912120952
$ l ~/tmp/16521
total 0
drwxrwxr-x. 3 akofink akofink 39 Sep 15 14:45 .
drwxr-xr-x. 4 akofink akofink 92 Sep 15 14:45 ..
-rw-rw-r--. 1 akofink akofink  0 Sep 15 14:45 bar
drwxrwxr-x. 2 akofink akofink 17 Sep 15 14:45 baz
-rw-rw-r--. 1 akofink akofink  0 Sep 15 14:45 foo
```

Again, just checking if this is intended behavior or if we should handle nested directories explicitly.

daviddavis:
Regarding pattern matching, I think that's the intended behavior. For example, if I specified something like `--path 'puppet-[0-9.-]{3,6}.rpm'`I wouldn't want the code to check each possible combination and report an error when each doesn't exist. We do however check if any files match at all and report an error if no files are uploaded.

For nested directories, when a directory is specified as `path`, it tries to upload all the files in the directory. This functionality has existed for a while so I'm leaning toward leaving it but if you think we should skip subdirs, or traverse them, or whatever, I can look into that (although I might suggest a follow up redmine issue depending on the work effort involved).

PR #452, opened 2016-09-20
Fixes #15978 - Fix host-collection content cmds

akofink:
Fix host-collection content commands:

```sh
hammer host-collection [package|package-group] [install|update|remove]
hammer host-collection erratum install
```

This change is needed due to host unification. The api changes slightly
to ensure bulk host operations go through the Foreman API rather than
the Katello api.

Prerequisite: https://github.com/Katello/katello/pull/6331


chris1984:
Tested and works

PR #453, opened 2016-09-23
Fixes #14290 - Remove unneeded require

akofink:
#14289 is no longer an issue, so this workaround is no longer necessary.


chris1984:
Ack looks good

chris1984:
Issue is resolved no longer need the require

PR #454, opened 2016-09-23
Fixes #16677 - Add LifecycleEnvironment list/create/update tests

akofink:
I'd like to at least have minimal options for most commands covered. Ideally, we'd verify that all options for each command are allowed, and the ones that are required are actually required.


thomasmckay:
ack

PR #455, opened 2016-09-27
Fixes #16708 - update endpoint for host collections

chris1984:


akofink:
It would be great if you would include a test for `hammer host-collection list --host <host_name>`

PR #456, opened 2016-09-29
Fixes #16730 - allow org options in repo delete

akofink:
allow organization options in hammer repository delete command


cfouant:
ACK!

PR #459, opened 2016-10-04
Fixes #16794 - Add org options to filter commands

daviddavis:
# TODO
- [x] Info test
- [x] Update test
- [x] Delete test
- [x] IdNameOptionsValidator test

### DIFF
--- a/lib/hammer_cli_katello/id_name_options_validator.rb
+++ b/lib/hammer_cli_katello/id_name_options_validator.rb
@@ -1,0 +1,73 @@
+module HammerCLIKatello
+  module IdNameOptionsValidator
+    # This is a method that requires:
+    # 1. either name or id has been supplied
+    # 2. if name is supplied, parent id/name/etc is also required
+    #
+    # Normally the parent will be organization as with products, content views,
+    # etc but this could also apply to repos for example whose names are unique
+    # per product
+    #
+    # Some examples:
+    #
+    # validate_id_or_name_with_parent
+    # validate_id_or_name_with_parent :content_view
+    # validate_id_or_name_with_parent :repository, parent: 'product', required: false
+    # validate_id_or_name_with_parent :content_view, parent: {organization: ['id', 'name']}
+    def validate_id_or_name_with_parent(record_name = nil, parent: 'organization', required: true)
+      child_options = IdNameOptionsValidator.build_child_options(record_name)
+      parent_options = IdNameOptionsValidator.build_parent_options(parent)
+
+      validate_options do
+        any(*child_options).required if required
+
+        if (name_option = child_options.detect { |opt| opt.end_with?('_name') }) &&
+           option(name_option).exist?
+          any(*parent_options).required
+        end
+      end
+    end
+
+    # This method simply checks that either id or name is supplied
+    #
+    # Some examples:
+    #
+    # # checks for a --id or --name option
+    # validate_id_or_name
+    # # checks for --content-view-id or --content-view-name
+    # validate_id_or_name :content_view
+    def validate_id_or_name(record_name = nil)
+      child_options = IdNameOptionsValidator.build_child_options(record_name)
+
+      validate_options do
+        any(*child_options).required
+      end
+    end
+
+    def self.build_child_options(record_name)
+      if record_name.nil?
+        %w(option_id option_name)
+      else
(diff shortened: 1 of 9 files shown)


akofink:
Another thing that could make the commands themselves better is to add validation so that the error messages are a bit more clear; all these commands require an organization if content view name is specified, and they all require some content view (whether by ID or resolved from a name).

Example: https://github.com/Katello/hammer-cli-katello/blob/60ed450e834583179447fcb203d274c9d49c6c34/lib/hammer_cli_katello/content_view_version.rb#L173

daviddavis:
I can add tests here for update/info/destroy. 

For validation, that's not unique to these commands (or even content view commands) correct? I am wondering if I maybe develop a reusable solution like a `NameOptionValidation` module or something. Thoughts?

akofink:
It'd definitely be good to dry it up. You're right that `--name` requires `--organization<-id|-label>` quite a bit. You might call it something that communicates the organization requirement, like `NameWithOrganizationOptionValidation` (I'm not sure about the exact name, of course).

daviddavis:
@akofink added a module to validate id/name and then if there's a name, check that parent id/name is also supplied. I updated list and info with the new module. Review it if you would and then if it's ok, I'll apply it to the other commands and then write some tests for it.

Also, note that I had to make a change to hammer-cli: https://github.com/theforeman/hammer-cli/pull/220

lib/hammer_cli_katello/id_name_options_validator.rb
```diff
@@ -0,0 +1,30 @@
...
+  # 1. either name or id is required
+  # 2. if name is supplied, some parent opt is also required
+  #
+  # Normally the parent will be organization as with products, content views,
+  # etc but this could also apply to repos for example whose names are unique
+  # per product
+
+  def validate_name_or_id_for(record_name, with_parent: 'organization', required: true)
+    validate_options do
+      parent_options = ["option_#{with_parent}_id", "option_#{with_parent}_name"]
```
akofink: I'm not sure this is flexible enough for the default case of organization because you can (always?) specify organization by `['option_organization_id', 'option_organization_name', 'option_organization_label']`. Perhaps you could specify the argument like this:

```ruby
validate_name_or_id_for(record_name, with_parent: {'organization' => ['name', 'id', 'label']}, required: true)
```

and allow more than just one parent option. For example, creating a host requires a myriad of other options (i.e. domain, operatingsystem, architecture, partition_table), maybe specified in this method like this:

```ruby
validate_name_or_id_for(record_name, with_parent: {'domain' => ['name', 'id'], 'operatingsystem' => ['name', 'id'], 'architecture' => ['name', 'id'], 'partition_table' => ['name', 'id']}, required: true)
```

I'm not convinced of these changes, however.

lib/hammer_cli_katello/id_name_options_validator.rb
```diff
@@ -0,0 +1,30 @@
...
+      record_options = if record_name.nil?
+                         %w(option_id option_name)
+                       else
+                         ["option_#{record_name}_id", "option_#{record_name}_name"]
+                       end
+
+      any(*record_options).required if required
+
+      if option(record_options.last).exist?
```
akofink: This would have to be modified if you have organization name or label.

test/functional/content_view/filter/list_test.rb
```diff
@@ -0,0 +1,64 @@
...
+FILTER ID | NAME | TYPE | INCLUSION
+----------|------|------|----------
+")
+
+    result = run_cmd(@cmd + params)
+    assert_cmd(expected_result, result)
+  end
+
+  it "lists content view filters" do
+    params = ["--organization-id=#{org_id}", "--content-view=#{cv_name}"]
```
akofink: Your tests don't cover the (currently failing) case of:

```sh
$ hammer content-view filter list --content-view 'cv14107' --organization-label 'Default_Organization'
Error: At least one of options --organization-id, --organization is required

See: 'hammer content-view filter list --help'
```

daviddavis:
@akofink with all these one off cases you bring up, I'm almost wondering if my idea to develop a DRY module is worth it. The module really doesn't encapsulate that much functionality. Maybe I should hold off at least until we write enough validate_options blocks that we can discern a pattern?

daviddavis:
@akofink updated (finally). should be ready for re-review.

PR #460, opened 2016-10-19
fixes #17031 - repository info: include mirror on sync

bbuckingham:


lib/hammer_cli_katello/repository.rb
```diff
@@ -75,6 +76,7 @@ class InfoCommand < HammerCLIKatello::InfoCommand
       def extend_data(data)
```
akofink: you may want to do a bit of refactoring on this method to reduce cyclomatic complexity for rubocop

bbuckingham: I am not sure if we can do too much to reduce this, since it is not a method that we define, but part of the general framework.

I would be tempted to either:
1. disable the cop on that specific method instance
2. not use the method

I prefer 1 as I really wanted to keep the behavior consistent with the other fields.

Thoughts?

akofink: I would try to separate some of this logic into a method, like `setup_sync_state` does.

PR #462, opened 2016-10-24
Fixes #17067 - Removing from environment options

daviddavis:
Seems like the from_lifecycle_environment option doesn't actually do anything as this command succeeds:

```
hammer content-view version promote --id 9 --from-lifecycle-environment ENVTHATDOESNOTEXIST --to-lifecycle-environment Dev --organization-id 1
```

I can even run promote without the option:

```
hammer content-view version promote --id 9 --to-lifecycle-environment Test --organization-id 1
```


akofink:
I'm not sure this command is broken. What about this command?

```
hammer content-view version promote --content-view-id 2 --organization-id 1 --to-lifecycle-environment development --from-lifecycle-environment Library
```

This is working for me on the master branch. I have an environment named 'development', and I'm using the 'Library' environment to find the content view version inside of the specified content view. If I leave off the `--from-lifecycle-environment` option, I get

```
Could not promote the content view:
  Error: found more than one content_view_version
```

daviddavis:
@akofink good catch. Closing.

PR #466, opened 2016-11-17
Fixes #17389 - Use hostgroup to fill in CV and LE

akofink:
This only happens if the user doesn't specify a content view or
lifecycle environment but does specify a hostgroup when creating a host.


cfouant:
@akofink - this may have already been fixed upstream? I am unable to reproduce this upstream, so I'm thinking it may have already been fixed. You might just need to link the redmine issue to whatever fixed this upstream It is reproducible in 6.2.2.

akofink:
Closing, as I can no longer reproduce the issue. Thanks @cfouant!

PR #467, opened 2016-12-09
Fixes #17619 - Provide correct org options in msg

akofink:
Provide the correct organization options to the user when trying to
resolve lifecycle environment name when creating/updating a hostgroup

### DIFF
--- a/lib/hammer_cli_katello/hostgroup_extensions.rb
+++ b/lib/hammer_cli_katello/hostgroup_extensions.rb
@@ -20,6 +20,16 @@ module HammerCLIKatello
         _('Organization name to search by'), attribute_name: :option_organization_name
       base.option '--query-organization-label', 'ORGANIZATION_LABEL',
         _('Organization label to search by'), attribute_name: :option_organization_label
+
+      base.validate_options do
+        organization_options = [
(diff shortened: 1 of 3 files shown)


lib/hammer_cli_katello/hostgroup_extensions.rb
```diff
@@ -20,6 +20,16 @@ def self.included(base)
         _('Organization name to search by'), attribute_name: :option_organization_name
       base.option '--query-organization-label', 'ORGANIZATION_LABEL',
         _('Organization label to search by'), attribute_name: :option_organization_label
+
+      base.validate_options do
+        organization_options = [
```
parthaa: can you reuse https://github.com/Katello/hammer-cli-katello/blob/master/lib/hammer_cli_katello/organization_options.rb#L12 here ?

akofink: The `OrganizationOptions` module isn't included here because the org options are `--query-organization[-id|-label]` rather than `--organization[-id|-label]`. In addition, this is an array of the  `attribute_name` of the options, not a hash of options.

parthaa:
ack

PR #469, opened 2016-12-13
Cherry picks 3.2.1

johnpmitsch:


akofink:
You will need https://github.com/Katello/hammer-cli-katello/pull/468 for passing tests.

ehelms:
@akofink even though it fixes tests, this branch is part of 1.13/3.2 where the hammer is older, I suspect if we keep the test fixing commit in here it will break in production

akofink:
@ehelms Yes, it must be paired with https://github.com/theforeman/hammer-cli/pull/227

ehelms:
I'd recommend pulling that comment and accepting that tests will not run on this branch.

johnpmitsch:
@ehelms its back to the 2 cherry picks and the version bump

PR #470, opened 2016-12-20
Fixes #17766 - Require org opts for product name

akofink:
Require organization options while listing errata if specifying a
product name.


johnpmitsch:
@akofink works as expected according the issue - I don't work in hammer much but code looks good.
ACK

PR #471, opened 2016-12-22
Fixes #17709 - Provide err msg for cv name in hg

akofink:
Provide proper error message to indicate required organization options
for content-view name in hostgroup create/update. These messages are
especially important for these commands because the organization options
are `--query-organization[-id|-label]` rather than just
`--organization[-id|-label]`.


thomasmckay:
ack thanks!

PR #472, opened 2017-01-13
Fixes #17691 - publish repo after all content uploads

jlsherrill:


jlsherrill:
Requires https://github.com/Katello/katello/pull/6544

lib/hammer_cli_katello/repository.rb
```diff
@@ -220,6 +221,8 @@ def execute
           File.open(file_path, 'rb') { |file| upload_file file }
```
akofink: The extra option to publish the repository could be given to the user or left out entirely, and this code here could be changed to upload the files much smarter with one call to `import_uploads` instead of N calls. [Your Katello code](https://github.com/Katello/katello/pull/6544) would then only publish once on the last upload rather than forcing `repository_publish: false` in hammer and then triggering a publish. It just seems like we have this logic in two places, and it's a non-ideal use of our API.

That said, this works great, and my above comment would be a performance improvement on top of the bug fix, so I would ACK this if you create an issue for cleaning up this code to only call import_upload once.

jlsherrill: I thought about that, but actually thought this might be better.  Doing it the way you suggest would mean that if an error occured part way through uploading the files, none of the files would be uploaded (because the import step wouldn't be called).  This way everything up to the error is fully uploaded.   I might could see arguments either way but i think i prefer this method.

akofink: Ok, that makes sense. In that case, ACK - Thanks @jlsherrill! btw, this PR helped a bunch to test your Katello PR.

PR #474, opened 2017-02-02
Fixes #18351 - Add org opts to some CV commands

akofink:
```
hammer content-view add-version
hammer content-view remove-version
hammer content-view copy
hammer content-view update
hammer content-view delete
```


bbuckingham:
ACK

PR #479, opened 2017-02-15
Fixes #18523 - hostgroup info is not showing content-related fields

upadhyeammit:
This adds 'Content Source' field to 'hammer hostgroup info' command.

### DIFF
--- a/test/functional/hostgroup/data/hostgroup.json
+++ b/test/functional/hostgroup/data/hostgroup.json
@@ -1,6 +1,6 @@
 {
-	"content_source_id":null,
-	"content_source_name":null,
+	"content_source_id":1,
+	"content_source_name":"foreman.example.com",
 	"content_view_id":1,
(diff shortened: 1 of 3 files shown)


akofink:
Works great. Could you please update this test to include content source? https://github.com/Katello/hammer-cli-katello/blob/master/test/functional/hostgroup/info_test.rb#L23

Let me know if you'd like some help (akofink on #theforeman-dev)

test/functional/hostgroup/data/hostgroup.json
```diff
@@ -1,6 +1,6 @@
 {
-	"content_source_id":null,
-	"content_source_name":null,
+	"content_source_id":1,
+	"content_source_name":foreman.example.com,
```
akofink: I think this line needs to be `"content_source_name":"foreman.example.com",`

PR #480, opened 2017-02-15
Fixes #18499 - Hammer CV filter rule accepts PackageGroup UUID

snagoor:
This is the second part of the fix from the Hammer CLI end which now accepts PackageGroup UUID as parameter. 

Also see https://github.com/Katello/katello/pull/6613

### DIFF
--- a/lib/hammer_cli_katello/package_group.rb
+++ b/lib/hammer_cli_katello/package_group.rb
@@ -5,7 +5,14 @@ module HammerCLIKatello
     class ListCommand < HammerCLIKatello::ListCommand
       output do
         field :id, _("ID")
-        field :name, _("Name")
+        field :name, _("Package Group Name")
+        field :repo_name, _("Repository Name")
+        field :uuid, _("UUID")
+      end
+
+      def extend_data(data)
+        data["repo_name"] = data["repository"]["name"]
+        data
       end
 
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/package_group.rb
```diff
@@ -5,7 +5,14 @@ class PackageGroupCommand < HammerCLIKatello::Command
     class ListCommand < HammerCLIKatello::ListCommand
       output do
         field :id, _("ID")
-        field :name, _("Name")
+        field :name, _("PackageGroup Name")
```
bbuckingham: Recommend: 

s/PackageGroup/Package Group/

lib/hammer_cli_katello/package_group.rb
```diff
@@ -16,14 +23,21 @@ class ListCommand < HammerCLIKatello::ListCommand
     class InfoCommand < HammerCLIKatello::InfoCommand
       output do
         field :id, _("ID")
-        field :name, _("Name")
+        field :name, _("PackageGroup Name")
```
bbuckingham: Recommend: 

s/PackageGroup/Package Group/

snagoor:
@bbuckingham Ack. Done the requested changes.

PR #485, opened 2017-02-21
Fixes #18589 - Modify docs in light of forklift

akofink:
Add more specific guides for testing. This PR is definitely open to comments for changes/additions to the documentation.


akofink:
@ehelms That's a great idea. I don't suspect the dev/test methods for hammer to change very much in the foreseeable future.

johnpmitsch:
@akofink looks good, my only suggestion is to show how to point the hammer devel box at a host, since  people who are not experienced with hammer will be wondering how to do that (like me :flushed:) and that is part of setting up the box (if I understand correctly).

akofink:
Good suggestion, @johnpmitsch! Added.

PR #491, opened 2017-03-09
Fixes #14611 - Add content-view purge command...

akofink:
Add content-view purge command to delete old versions of a content view. Thanks to @rclsilver for the work to get this started.

### DIFF
--- a/lib/hammer_cli_katello/content_view_purge.rb
+++ b/lib/hammer_cli_katello/content_view_purge.rb
@@ -1,0 +1,93 @@
+module HammerCLIKatello
+  class ContentViewPurgeCommand < HammerCLIKatello::Command
+    include HammerCLIForemanTasks::Async
+    include OrganizationOptions
+
+    command_name "purge"
+    desc "Delete old versions of a content view"
+
+    option "--id", "ID", _("Content View numeric identifier")
+    option "--name", "NAME", _("Content View name")
+    option "--count", "COUNT", _("count of unused versions to keep"),
+           default: 3, format: HammerCLI::Options::Normalizers::Number.new
+
+    validate_options do
+      organization_options = [:option_organization_id, :option_organization_name, \
+                              :option_organization_label]
+
+      any(:option_name, :option_id).required
+
+      if option(:option_name).exist?
+        any(*organization_options).required
+      end
+    end
+
+    build_options
+
+    def resource_content_views
+      HammerCLIForeman.foreman_resource(:content_views)
+    end
+
+    def resource_content_view_versions
+      HammerCLIForeman.foreman_resource(:content_view_versions)
+    end
+
+    def all_options
+      result = super.clone
+      if result['option_id'].nil? && result['option_name']
+        result['option_id'] = resolver.content_view_id(result)
+      end
+      @all_options = result
+    end
+
+    def execute
+      if option_count < 0
+        output.print_error _("Invalid value for --count option: value must be 0 or greater.")
+        return HammerCLI::EX_USAGE
+      end
+
+      # Check if there is something to do
+      if option_count >= old_unused_versions.size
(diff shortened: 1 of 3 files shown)


lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -0,0 +1,54 @@
...
+        end
+
+        # Keep only versions which are not used
+        old_versions = all_versions.select do |v|
+          v["environment_ids"].empty?
+        end
+
+        # Check if there is something to do
+        if option_count >= old_versions.size
+          output.print_error _("No version to delete.")
```
bbuckingham: s/version/versions/

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -0,0 +1,54 @@
...
+          output.print_error _("No versions to delete.")
+          HammerCLI::EX_NOT_FOUND
+        else
+          versions_to_purge = old_versions.slice(0, old_versions.size - option_count)
+
+          versions_to_purge.each do |v|
+            HammerCLIForeman.record_to_common_format(
+              resource_content_view_versions.call(:destroy, 'id' => v["id"])
+            )
+            print_message _("Version '%{version}' of content view '%{view}' deleted.") %
```
bbuckingham: Since the destroy call is an async operation, I wonder if we should indicate that it has been scheduled vs deleted?

In the future, it may be useful to enhance it so that it monitors the destroy through completion.  (e.g. using foreman tasks)

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -0,0 +1,54 @@
+module HammerCLIKatello
+  class ContentViewPurgeCommand < HammerCLIKatello::Command
+    command_name "purge"
+    desc "Delete old versions of a content view"
+
+    option "--id", "ID", _("content view numeric identifier"), required: true
```
bbuckingham: For consistency, we may want to enhance this in the future to also support --name, which would mean id resolving with organization/organization-id.  We can consider this outside of the scope of the PR, but may be useful to create a redmine to track.

test/functional/content_view/purge_test.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    it 'fails gracefully when there are no versions to delete' do
+      ex = api_expects(:content_views, :show) do |p|
+        p['id'] == '2'
+      end
+      ex.returns(versions)
+
+      r = run_cmd(%w(content-view purge --id 2 --count 3))
+      assert(r.err.include?('No version to delete.'), 'Incorrect error message')
```
bbuckingham: s/version/versions/

bbuckingham:
Thanks @rclsilver and @akofink !   Tested on my environment and it worked well.  Only a few minor comments and I am ok with tracking future enhancements with additional redmines.

akofink:
@bbuckingham I went ahead and added the enhancements mentioned, as they were supported by existing helpers and didn't pose much effort.

bbuckingham:
Uh oh... Just observed the purge delete 3 unexpected content view versions and the output is missing for version and view.

```
hammer> content-view purge --organization "Default Organization" --name zoo --count 1
[.................................................................................................................................................................................] [100%]

Version '' of content view '' deleted.
[.................................................................................................................................................................................] [100%]

Version '' of content view '' deleted.
[.................................................................................................................................................................................] [100%]

Version '' of content view '' deleted.
[.................................................................................................................................................................................] [100%]

Version '' of content view '' deleted.
[.................................................................................................................................................................................] [100%]

Version '' of content view '' deleted.

```

bbuckingham:
ACK.  The output is correctly showing version and cv now.

The count was a non-issue and behaving as designed.  :)

PR #494, opened 2017-03-29
Fixes #19079 - Update options overrides

akofink:
These changes are to handle the new changes in hammer-cli introduced in
https://github.com/theforeman/hammer-cli/pull/233

### DIFF
--- a/lib/hammer_cli_katello/composite_content_view_name_resolvable.rb
+++ b/lib/hammer_cli_katello/composite_content_view_name_resolvable.rb
@@ -10,15 +10,13 @@ module HammerCLIKatello
       }
     end
 
-    def all_options
-      result = super.clone
-      if result['option_composite_content_view_name'] &&
-         result['option_composite_content_view_id'].nil?
-        options = composite_content_view_resolve_options(result)
-        result['option_composite_content_view_id'] = resolver.content_view_id(options)
-        @all_options = result
+    def options
+      if super['option_composite_content_view_name'] &&
+         super['option_composite_content_view_id'].nil?
+        options = composite_content_view_resolve_options(super)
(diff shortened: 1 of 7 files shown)


lib/hammer_cli_katello/composite_content_view_name_resolvable.rb
```diff
@@ -10,15 +10,13 @@ def composite_content_view_resolve_options(options)
...
 
-    def all_options
-      result = super.clone
-      if result['option_composite_content_view_name'] &&
-         result['option_composite_content_view_id'].nil?
-        options = composite_content_view_resolve_options(result)
-        result['option_composite_content_view_id'] = resolver.content_view_id(options)
-        @all_options = result
+    def options
+      if super['option_composite_content_view_name'] &&
```
parthaa: is better to save super in a variable like result and set that ?

akofink: I'm trying to modify the instance variable @options or @all_options in the [option_collector](https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/options/option_collector.rb#L12). The methods use the `||=` syntax to cache the result, so it's not bad performance-wise.

parthaa:
@akofink getting this error with your changes. I verified the same works fine in master
https://paste.fedoraproject.org/paste/cS-3NKgrIm7jo-CEzG5LxV5M1UNdIGYhyRLivL9gydE=

parthaa:
@akofink basically try ```hammer -vd content-view component list --composite-content-view=<Composite Content View Name> --organization-id=1``` and make sure that works

parthaa:
n/m it worked after I ```rm Gemfile.lock ; bundle install```

ACK

PR #496, opened 2017-04-04
Fixes #18923 - Allow ID in puppet module remove

akofink:
Allow passing puppet module ID *or* UUID to remove a puppet module from
a content view


omaciel:
@akofink If cross-referencing across different organizations works in Github, then mentioning  @SatelliteQE/quality-engineers should do the trick to inform QE about important changes

omaciel:
Hmmm seems that no notification was sent :/

thomasmckay:
ack

PR #497, opened 2017-04-11
Fixes #19184 - Test commands used in sat-clone

akofink:
Issues: [#19184](http://projects.theforeman.org/issues/19184)


johnpmitsch:
ACK (but know that I'm not too familiar with this area of code) :smile:

PR #498, opened 2017-04-17
Fixes #19262 - Return 1 if any service fails ping

akofink:


lib/hammer_cli_katello/ping.rb
```diff
@@ -44,7 +44,8 @@ class PingCommand < HammerCLIKatello::Command
     def execute
       d = send_request
       print_data d
-      d['status'] != _("FAIL") ? HammerCLI::EX_OK : 1
+      service_failure = d['services'].values.map {|v| v['status'] }.include?(_("FAIL"))
```
ehelms: Could could use any? here to make this more concise

akofink: Could you please elaborate? I'm having trouble seeing how `any?` would work in this situation.

lib/hammer_cli_katello/ping.rb
```diff
@@ -44,7 +44,8 @@ class PingCommand < HammerCLIKatello::Command
     def execute
       d = send_request
       print_data d
-      d['status'] != _("FAIL") ? HammerCLI::EX_OK : 1
+      service_failure = d['services'].values.map { |v| v['status'] }.include?(_("FAIL"))
+      service_failure || d['status'] != _("FAIL") ? 1 : HammerCLI::EX_OK
```
johnpmitsch: It looks like the `d['status'] != _("FAIL")` should be `d['status'] == _("FAIL") ` as that will return 1 if true and 0 if false.

And this is up to you, but this one-liner is a bit hard to parse, it would be nice if it was broken up

johnpmitsch: a service being down returned the correct exit code, but the overall status did not:
```
[vagrant@centos7-hammer-devel hammer-cli-katello]$ hammer ping
candlepin:      
    Status:          ok
    Server Response: Duration: 48ms
candlepin_auth: 
    Status:          ok
    Server Response: Duration: 50ms
pulp:           
    Status:          ok
    Server Response: Duration: 61ms
foreman_tasks:  
    Status:          ok
    Server Response: Duration: 47ms

{"status"=>"FAIL", "services"=>{"pulp"=>{"status"=>"ok", "duration_ms"=>"61", "_response"=>"Duration: 61ms"}, "pulp_auth"=>{"_response"=>nil}, "candlepin"=>{"status"=>"ok", "duration_ms"=>"48", "_response"=>"Duration: 48ms"}, "candlepin_auth"=>{"status"=>"ok", "duration_ms"=>"50", "_response"=>"Duration: 50ms"}, "foreman_tasks"=>{"status"=>"ok", "duration_ms"=>"47", "_response"=>"Duration: 47ms"}}}
[vagrant@centos7-hammer-devel hammer-cli-katello]$ echo $?
0
```

akofink:
Waiting on you, @ehelms

PR #504, opened 2017-08-16
Fixes #20613 - Update help text for --sync-date

ahumbe:
Current help text for "hammer sync-plan create" is confusing for --sync-date option. 

Changed the text 
From: 
"start date and time of the synchronization defaults to now"
To: 
"Start date and time for the sync plan. Time is optional, if kept blank current system time will be considered"

Sync plan get created even if we do not pass the time in "YYYY-MM-DD HH:MM:SS" format, date is mandatory.

$ hammer  sync-plan create --organization-id 1 --name "RedHat Software collections sync" --description "A sync plan for RHSCL" --enabled True --interval "weekly" --sync-date 2017-08-16
Sync plan created

### DIFF
--- a/lib/hammer_cli_katello/sync_plan.rb
+++ b/lib/hammer_cli_katello/sync_plan.rb
@@ -31,7 +31,8 @@ module HammerCLIKatello
              )
 
       option "--sync-date", "SYNC_DATE",
-             _("start date and time of the synchronization defaults to now"),
+             _("Start date and time for the sync plan." \
+               "Time is optional, if kept blank current system time will be considered"),
              :format => HammerCLI::Options::Normalizers::DateTime.new
 
       success_message _("Sync plan created")


ahumbe:
After this change the help look like: 

 --sync-date SYNC_DATE                   Start date and time for the sync plan. Time is optional, if kept blank current system time will be considered
                                         Date and time in YYYY-MM-DD HH:MM:SS or ISO 8601 format

lib/hammer_cli_katello/sync_plan.rb
```diff
@@ -31,7 +31,8 @@ class CreateCommand < HammerCLIKatello::CreateCommand
              )
 
       option "--sync-date", "SYNC_DATE",
-             _("start date and time of the synchronization defaults to now"),
+             _("Start date and time for the sync plan." \
+               "Time is optional, if kept blank current system time will be considered"),
```
akofink: Hmm... actually, time is not optional (via the API).

ahumbe: Yeah, sync-date is not optional. We can not create sync plan without sync date. Time part of the syntax  "YYYY-MM-DD HH:MM:SS" is optional, so we can just pass the date "YYYY-MM-DD".

Is this change acceptable ?

akofink: Ah, I see. That makes sense.

PR #510, opened 2017-09-22
0.11.1 cherry-picks

akofink:


akofink:
This needs up to date API cache for the tests. These should be generated from Katello 3.5.0.rc1

PR #530, opened 2017-11-21
fixes #21275 - unique identifier needed for content-override name

cfouant:
Requires https://github.com/Katello/katello/pull/7081

@parthaa if you don't mind being a reviewer on this as well...

### DIFF
--- a/lib/hammer_cli_katello/content_override.rb
+++ b/lib/hammer_cli_katello/content_override.rb
@@ -11,10 +11,10 @@ module HammerCLIKatello
         option "--content-label", "CONTENT_LABEL", _("Label of the content"),
                :attribute_name => :option_content_label, :required => true
 
-        option "--name", "NAME", _("Name/Key to override. " \
-                                   "To enable or disable a repo select 'enabled'." \
+        option "--override-name", "OVERRIDE_NAME", _("Override parameter key or name.\n" \
+                                   "To enable or disable a repo select 'enabled'.\n" \
                                    "Default value: enabled"),
-               :attribute_name => :option_name, :default => "enabled"
+               :attribute_name => :option_override_name, :default => "enabled"
 
         option "--value", "VALUE", _("Override value. " \
                                      "Note for repo enablement you can use a boolean value"),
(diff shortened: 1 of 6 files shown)


akofink:
Actually, I just realized, the name option should be the activation key name for ID resolution. That's basically why this bug was reported (it seems to me) because the user was confused about the name option, which behaves differently than any other hammer command.

That being said, the API never sees this activation key name param because we resolve the AK ID from it and pass that to the API. We could fix this bug simply by (not changing the API) changing the options in this command to `--name` for the AK name and something like `--override-name` for the name API param. Just a suggestion for discussion, not a request.

cfouant:
Since the API hasn't actually hit any Katello release, upstream or down, Partha and I discussed this and thought the better solution would be to fix the API.

cfouant:
@akofink Since the API hasn't actually hit any Katello release, upstream or down, Partha and I discussed this and thought the better solution would be to fix the API.

lib/hammer_cli_katello/content_override.rb
```diff
@@ -39,7 +39,7 @@ def request_params
           override = { 'content_label' => option_content_label }
           override['value'] = option_value if option_value
           override['remove'] = true if option_remove?
-          override['name'] = option_name
+          override['override_name'] = option_override_name
```
parthaa: This change is unnecessary imo. Cant you just do
```
override['name'] = option_override_name
```
and get away with it. So you keep the ```--override-name``` option but not change the API

parthaa:
@akofink @cfouant might want to address my comment https://github.com/Katello/hammer-cli-katello/pull/530#pullrequestreview-79270145 before merging.. My suggestion would not require any change in the API . Just change the name of the flag and  update the request_params method in hammer.

akofink:
@parthaa I agree with you that that would be the ideal change. I don't see the `name` param as a problem considering it's nested under `content_overrides`. If I had to vote, I'd vote to not change the param, but I don't feel strongly about it.

cfouant:
@parthaa @akofink changes made in agreement! Please re-review and let me know if you would like any additional changes, thanks!

parthaa:
ACK

PR #541, opened 2018-02-02
Fixes #22504 Proper-ize the error expectations

akofink:
Includes https://github.com/Katello/hammer-cli-katello/pull/540 for testing purposes. It should be green.


zjhuntin:
ACK

PR #542, opened 2018-02-07
Fixes #22517 - Resolvables compatible with option sources

mbacovsky:
Redefinitions of all_options were converted to option_sources. The reason for the change is that with recent introduction of option_sources all resolving should happen there. Option sources are evaluated before all_options and thus overriding resolvers in all_options no longer work.  

I've added some refactoring in separate commit. It is removing duplicite code. Please review the changes there carefully especially if this duplicites were intentional.

I've also found one more occurrence of all_options here https://github.com/Katello/hammer-cli-katello/blob/0e31b71ef0681bd57543f42fd075470bf9a9abe5/lib/hammer_cli_katello/file.rb#L61
but I don't understand why the method is there and what it does. From my testing it seems it is not called and tests pass when I delete it. Could it be removed?


akofink:
The `all_options` in `file.rb` can be removed safely. That code is well covered by tests.

mbacovsky:
@akofink, cool, updated

mbacovsky:
@akofink the related hammer-cli-foreman PR was merged. Do you still see the 2 failures?

akofink:
No failures! The previous failures were caused by running hammer-cli-foreman-tasks on master.

PR #552, opened 2018-04-19
Fixes #23323 - Shows ignorable units in repo

parthaa:
This commit feeds of the upstream 'ignorable content' feature and shows
content units that are going to be ignored during a sync operation for
that repo. This is only available for yum repos.

### DIFF
--- a/lib/hammer_cli_katello/repository.rb
+++ b/lib/hammer_cli_katello/repository.rb
@@ -44,6 +44,7 @@ module HammerCLIKatello
               Fields::Field, :hide_blank => true
         field :container_repository_name, _("Container Repository Name"),
               Fields::Field, :hide_blank => true
+        field :ignorable_content, _("Ignorable Content Units"), Fields::List, :hide_blank => true
 
         label _("Product") do
           from :product do


parthaa:
blocked by https://github.com/Katello/katello/pull/7325

parthaa:
```
$ hammer repository info --id=69
ID:                      69
Name:                    repo
Label:                   repo
...............
Download Policy:         on_demand
Ignorable Content Units: srpm
..............
```

parthaa:
[test]

akofink:
Rebase for green tests! :)

lib/hammer_cli_katello/repository.rb
```diff
@@ -83,10 +84,14 @@ class InfoCommand < HammerCLIKatello::InfoCommand
       end
 
       def extend_data(data)
-        if data["content_type"] == "yum" && data["gpg_key"]
-          data["gpg_key_name"] = data["gpg_key"]["name"]
+        if data["content_type"] == "yum"
+          if data["ignorable_content"]
+            data["_ignorable_content"] = data["ignorable_content"].join(",")
```
akofink: I don't mind overwriting the data coming back from the API. No need to make a new key imo. Also, [hammer-cli-foreman does it](https://github.com/theforeman/hammer-cli-foreman/blob/170e7d8aac5b7d05d76ce0379cfde4ff2de38f65/lib/hammer_cli_foreman/settings.rb#L18).

parthaa: > It looks good, just one comment. I'd like you to include an updated apipie cache since you changed the API in Katello.

Not needed. I see the updated API json merged already .

akofink:
Looks like the master branch has the change in the apipie cache afterall, so no need to regenerate.

PR #554, opened 2018-05-03
Fixes #23484: Test message format in tests

akofink:
Basically copied https://github.com/theforeman/hammer-cli-foreman/pull/352. I believe I have 8 strings to update before this PR is green.


sjha4:
Changes look good based on the test definition.
**** 
Just curious, why is it that the message format tests want desc and error msgs to not end with a period but status msg to end with a period?

akofink:
@sjha4 Did you want to ACK this?

sjha4:
@akofink : Yes :+1: ACK

PR #563, opened 2018-06-18
Fixes #22696 - Fix errors when --packages is used

shiramax:


akofink:
This issue stems from the fact that the [packages API](https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/packages_controller.rb#L40) uses `search` instead of the name parameter directly (i.e. see the [products controller](https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/products_controller.rb#L35) as an example). This is the way that all of core Foreman's index APIs are, but Katello has a mixed API, unfortunately (believe me, there is always ranting happening about this, but it's a lot of work to unify the search pattern).

Anyway, the easiest way to handle this is to use the `lib/hammer_cli_katello/foreman_search_options_creators.rb`. This will tell hammer-cli-katello to treat packages like a Foreman API resource for searching.

shiramax:
@akofink thanks :) please review again .

shiramax:
@akofink Done, please review again :)

test/functional/content_view/version/incremental_update_test.rb
```diff
@@ -88,4 +88,18 @@
     result = run_cmd(@cmd + params)
     assert_equal(result.exit_code, 0)
   end
+
+  it "performs incremental update with packages" do
+    params = ['--content-view-version-id=5', '--package-ids=1', '--lifecycle-environment-ids=1']
```
akofink: We want to test the `--packages` option specifically rather than `--package-ids`. This will create another request to `packages#index`.

shiramax:
@akofink, thanks.
please review again :)

test/functional/content_view/version/incremental_update_test.rb
```diff
@@ -88,4 +88,21 @@
...
+
+  it "performs incremental update with packages" do
+    params = ['--content-view-version-id=5', '--packages=bla', '--lifecycle-environment-ids=1']
+
+    expect_generic_search(
+      :packages, params: {search: "name = \"bla\""}, returns: {'id' => 5}
+    ).at_least_once
+
+    ex = api_expects(:content_view_versions, :incremental_update, 'Incremental Update') do |par|
+      par[:content_view_version_environments][0][:content_view_version_id] == 5
```
akofink: The package ids returned from the previous search should be passed in here.

test/functional/content_view/version/incremental_update_test.rb
```diff
@@ -88,4 +88,21 @@
     result = run_cmd(@cmd + params)
     assert_equal(result.exit_code, 0)
   end
+
+  it "performs incremental update with packages" do
+    params = ['--content-view-version-id=5', '--packages=bla', '--lifecycle-environment-ids=1']
+
+    expect_generic_search(
+      :packages, params: {search: "name = \"bla\""}, returns: {'id' => 5}
+    ).at_least_once
```
akofink: `at_least_once` allows for multiple searches. It should work without it (thus expecting only one call).

shiramax:
@akofink thanks!
please review again :)

test/functional/content_view/version/incremental_update_test.rb
```diff
@@ -88,4 +88,21 @@
...
+
+    expect_generic_search(
+      :packages, params: {search: "name = \"bla\""}, returns: {'id' => 5}
+    )
+
+    ex = api_expects(:content_view_versions, :incremental_update, 'Incremental Update') do |par|
+      par[:content_view_version_environments][0][:content_view_version_id] == 5 &&
+        par["add_content"]["package_ids"] == [5]
+    end
+    ex.returns(index_response([{'id' => 5}]), 'id' => '3', 'state' => 'stopped')
```
akofink: The index_reponse should be on line 96, not here. You only need the `'id' => '3'` here.

test/functional/content_view/version/incremental_update_test.rb
```diff
@@ -88,4 +88,21 @@
     result = run_cmd(@cmd + params)
     assert_equal(result.exit_code, 0)
   end
+
+  it "performs incremental update with packages" do
+    params = ['--content-view-version-id=5', '--packages=bla', '--lifecycle-environment-ids=1']
+
+    expect_generic_search(
+      :packages, params: {search: "name = \"bla\""}, returns: {'id' => 5}
```
akofink: This test would be more clear if the package ID didn't match the content_view_version_id. Just pick a different number.

shiramax:
@akofink thanks ! done

test/functional/content_view/version/incremental_update_test.rb
```diff
@@ -88,4 +88,21 @@
     result = run_cmd(@cmd + params)
     assert_equal(result.exit_code, 0)
   end
+
+  it "performs incremental update with packages" do
+    params = ['--content-view-version-id=5', '--packages=bla', '--lifecycle-environment-ids=1']
+
+    expect_generic_search(
+      :packages, params: {search: "name = \"bla\""}, returns: {'id' => 15}
+    ).returns(index_response([{'id' => 15}]))
```
akofink: You can put the returns here or on line 96, but you don't need it in both places. I think the last one gets honored, so the `returns: {'id' => 15}` isn't doing anything.

shiramax:
@akofink done :)

PR #564, opened 2018-06-20
fixes #24023 - removes singular name option if array of names exists

cfouant:


lib/hammer_cli_katello/search_options_creators.rb
```diff
@@ -20,8 +20,11 @@ def create_repositories_search_options(options, _mode = nil)
       product_id = options[HammerCLI.option_accessor_name("product_id")]
 
       search_options = {}
-      search_options['name'] = name if name
-      search_options['names'] = names if names
+      if names
+        search_options['names'] = names
+      elsif name
+        search_options['name'] = name
+      end
```
akofink: @tstrachota A side effect of [this change](https://github.com/theforeman/hammer-cli-foreman/pull/346/files#diff-25c31437e73b23714ac3ea197aa60b15R142):

```rb
all_opts = {"option_organization_id"=>1,
            "option_repository_names"=>["zoo"],
            "option_name"=>"cv2", # content view name, irrelevant for resolving repository IDs
            "option_product_id"=>2}

scoped_options("repository", all_opts, :multi)
# {"option_organization_id"=>1,
#  "option_name"=>"cv2",
#  "option_product_id"=>2,
#  "option_names"=>["zoo"]}

scoped_options("repository", all_opts) # mode = nil, both :multi and :single (how it worked before)
# {"option_organization_id"=>1,
#  "option_product_id"=>2,
#  "option_names"=>["zoo"]}
```

Because the "option_repository_name" key doesn't exist in `all_opts`, "option_name" gets passed through `scoped_options` unchanged and requires logic here to ignore it and prefer "names", otherwise pass the content view name to the repositories search API, resulting in this bug.

Is this a good place to fix this? Are we doing something non-kosher with this search option creator that's causing it to be different than the rest?

tstrachota: I think it's fine.
We really need to redesign the whole resolving machinery to make it more transparent :-/

PR #566, opened 2018-06-25
Fixes #24060 - fix Puppet environments for host commands

iNecas:
Before this patch, the --environment option in host command was trying
to search in lifecycle environments. Reusing the resolvable definition
from the hostgroup, as they both should have the same behavior.


iNecas:
No idea, maybe @mbacovsky would know, but he's out for two weeks, so I would probably not hold this PR before he comes back.

akofink:
I made this issue: https://projects.theforeman.org/issues/24063

PR #570, opened 2018-07-13
Fixes #24249 - fix content-upload functionality

iNecas:
1. decrease the chunk size to be compatible with Pulp's expectations
2. suppress false-positive warnings from rest-client
3. log exceptions properly

### DIFF
--- a/lib/hammer_cli_katello/repository.rb
+++ b/lib/hammer_cli_katello/repository.rb
@@ -238,8 +238,9 @@ module HammerCLIKatello
           }
         ], last_file: true)
         print_message _("Repository updated")
-      rescue
+      rescue => e
         @failure = true
+        logger.error e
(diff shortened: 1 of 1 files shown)


lib/hammer_cli_katello/repository.rb
```diff
@@ -441,6 +447,16 @@ def print_results(name, results)
           end
         end
       end
+
+      def silence_warnings
+        original_verbose = $VERBOSE
```
akofink: where does $VERBOSE come from?

iNecas: Directly in Ruby, see https://mislav.net/2011/06/ruby-verbose-mode/ for some more details. Basically, it suppresses the output from `warn` calls.

iNecas:
Any chance to merge this then?

PR #573, opened 2018-08-02
Fixes #24528: s/smart_proxy/Content Source/

akofink:
When users provide `--content-source foo`, they should receive back
error messages talking about content source, not smart proxies.

### DIFF
--- a/lib/hammer_cli_katello/host_content_source_options.rb
+++ b/lib/hammer_cli_katello/host_content_source_options.rb
@@ -20,6 +20,9 @@ module HammerCLIKatello
           resource_hash["content_source_id"] = resolver.smart_proxy_id(proxy_options)
         end
       end
+    rescue HammerCLIForeman::ResolverError => e
+      e.message.gsub!('smart_proxy', _('Content Source'))
+      raise e
     end
   end
 end


stbenjam:
I still get the SQL error, also I think you have the wrong redmine issue linked? https://projects.theforeman.org/issues/24523


```
[vagrant@centos7-hammer-devel ~]$ hammer -u admin -p changeme -s https://centos7-devel.astriaporta.example.com/ hostgroup create --name potato --content-source-id 92
Could not create the hostgroup:
  PG::ForeignKeyViolation: ERROR:  insert or update on table "hostgroups" violates foreign key constraint "hostgroups_content_source_id_fk"
  DETAIL:  Key (content_source_id)=(92) is not present in table "smart_proxies".
  : INSERT INTO "hostgroups" ("name", "created_at", "updated_at", "title", "content_source_id", "lookup_value_matcher") VALUES ($1, $2, $3, $4, $5, $6) RETURNING "id"
```

akofink:
One number off! ;) 24528

akofink:
Doh! So, the first issue on this BZ was to fix that error. I did that back when I thought both host and hostgroup had a content_facet, so it looks like it's only fixed for hosts. :/

akofink:
@stbenjam https://github.com/Katello/katello/pull/7608 should fix the SQL error for hostgroups.

PR #575, opened 2018-08-10
Fixes #24582 - search_options supports plural resolution

ofedoren:


ofedoren:
@akofink, could you please take a look?

ofedoren:
@akofink, yes, while I was writing hammer ansible plugin I assumed that the plural resolution names->ids was working (see: https://github.com/theforeman/hammer-cli-foreman/blob/cef2d066ead6789dd08656466c8d614463b948b1/lib/hammer_cli_foreman/id_resolver.rb#L329). But I guess because of https://github.com/Katello/hammer-cli-katello/blob/31e2fb825d4ec563cc069850a785f1e79f0a31cf/lib/hammer_cli_katello/id_resolver.rb#L50 that method was "replaced" with this one from katello, which caused an error on plural resolution.

akofink:
Which resources specifically? If they're Foreman resources, use `create_search_options_without_katello_api`.

ofedoren:
@akofink yeah... I saw that, but I'm not sure I can add something [similar](https://github.com/Katello/hammer-cli-katello/blob/1dc05861ed2121a5f02db6f70efdb367bf4cd61b/lib/hammer_cli_katello/foreman_search_options_creators.rb#L20) with `ansible_roles` instead of `hosts` here because the `ansible_roles` resource comes from Foreman's [Ansible plugin](https://github.com/theforeman/foreman_ansible/blob/ec84e8b201ef4a9179c356608e930f1f0f58f5c6/app/controllers/foreman_ansible/api/v2/hosts_param_group_extensions.rb#L8) and adds to `host create/update` commands the `--ansible-roles` option, so in the time of names->ids resolving the HammerCLIForeman::IdResolver is used. but with `create_search_options_with_katello_api` method from HammerCLIKatello::IdResolver. And I don't really want to add another one IdResolver which will combine? mentioned two into the HammerCLIForemanAnsible::IdResolver.

I hope you can understand my problem, if not I'm ofedoren on the irc.

lib/hammer_cli_katello/search_options_creators.rb
```diff
@@ -87,12 +87,15 @@ def create_host_collections_search_options(options, mode = nil)
...
-    def create_search_options_with_katello_api(options, resource, _mode = nil)
+    def create_search_options_with_katello_api(options, resource, mode = nil)
       search_options = {}
       searchables(resource).each do |s|
         value = options[HammerCLI.option_accessor_name(s.name.to_s)]
-        if value
+        values = options[HammerCLI.option_accessor_name(s.plural_name.to_s)]
+        if value && [:single, nil].include?(mode)
           search_options.update(s.name.to_s => value.to_s)
+        elsif values && [:multi, nil].include?(mode)
```
akofink: Don't include nil here. It'll never make it as it's short circuited on L95.

akofink:
I don't think I understand. Shouldn't the following work to resolve the --ansible-roles-ids?

```rb
  def create_ansible_roles_search_options(options, mode = :multi)
    create_search_options_without_katello_api(options, api.resource(:ansible_roles), mode)
  end
```

ofedoren:
@akofink, sure it'll work if I add it. The problem is I'm not sure where to add it:
1. Having it in hammer-cli-katello plugin doesn't make sense since `ansible_roles` comes from Ansible plugin.
2. Having it in hammer-cli-foreman doesn't make sense as well because there is no need at all.
3. Having it in hammer-cli-foreman-ansible plugin requires to add hammer-cli-katello as a depency to use your solution (since katello is not a depency of foreman-ansible I don't think hammer plugin should be).

Please correct me if I'm mistaken.

akofink:
I see now, thank you for the explanation. I was also concerned about this, and I agree with all three points you made. :)

lib/hammer_cli_katello/search_options_creators.rb
ofedoren: What if value == nil && values != nil && mode == nil?

akofink: Oh, sorry :) I see now this is values, not value.

PR #578, opened 2018-08-29
Bump to 0.14.0

akofink:


sjha4:
Don't see the version bump in gemspec but I don't know enough here :+1: ACK!

akofink:
It's already bumped. We bump it during branching (I'm about to bump master to 0.15.0).

ekohl:
It's probably good to keep in the version constraints but change `~>` to `>=` before releasing. Otherwise I'll nag about it in packaging ;)

PR #581, opened 2018-09-04
Fixes #24806: bound h-c-f version in gemspec

akofink:


hammer_cli_katello.gemspec
```diff
@@ -47,7 +47,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ['lib']
   gem.version = HammerCLIKatello.version
 
-  gem.add_dependency 'hammer_cli_foreman'
+  gem.add_dependency 'hammer_cli_foreman', '~> 0.14.0'
```
ehelms: Travis tests don't like this. I think it was previously just `'~> 0.14'`

PR #584, opened 2018-09-11
Fixes #24894 - Support for cron expressions

sjha4:
Adds support for custom cron expressions in sync plans.

Requires https://github.com/Katello/katello/pull/7644 , https://github.com/Katello/katello/pull/7700

### DIFF
--- a/lib/hammer_cli_katello/sync_plan.rb
+++ b/lib/hammer_cli_katello/sync_plan.rb
@@ -9,6 +9,7 @@ module HammerCLIKatello
         field :sync_date, _("Start Date"), Fields::Date
         field :interval, _("Interval")
         field :enabled, _("Enabled"), Fields::Boolean
+        field :cron_expression, _("Cron Expression")
       end
 
       build_options
@@ -30,10 +31,9 @@ module HammerCLIKatello
     end
 
     class CreateCommand < HammerCLIKatello::CreateCommand
-      option "--interval", "INTERVAL", _("how often synchronization should run"),
-             :format => HammerCLI::Options::Normalizers::Enum.new(
-               %w(hourly daily weekly)
-             )
+      option "--interval", "INTERVAL", _("how often synchronization should run")
+
+      option "--cron-expression", "CRON EXPRESSION", _("set this when interval is custom cron")
 
       option "--sync-date", "SYNC_DATE",
              _("Start date and time for the sync plan." \
@@ -47,10 +47,8 @@ module HammerCLIKatello
     end
 
     class UpdateCommand < HammerCLIKatello::UpdateCommand
-      option "--interval", "INTERVAL", _("how often synchronization should run"),
-             :format => HammerCLI::Options::Normalizers::Enum.new(
-               %w(hourly daily weekly)
-             )
+      option "--interval", "INTERVAL", _("how often synchronization should run")
+
       option "--sync-date", "SYNC_DATE", _("start date and time of the synchronization"),
              :format => HammerCLI::Options::Normalizers::DateTime.new
 
--- a/test/functional/sync_plan/create_test.rb
+++ b/test/functional/sync_plan/create_test.rb
@@ -1,0 +1,60 @@
+require_relative '../test_helper'
+
+describe "create sync plan" do
+  let(:org_id) { 1 }
+  let(:name) { "sync_plan1" }
+  let(:hourly) { "hourly" }
+  let(:custom) { "custom cron" }
+  let(:cron) { "10 * * * *" }
+  let(:date) { "2018-09-08T00:00:00+00:00" }
+
+  it 'with organization ID,name,interval,date and enabled' do
+    api_expects(:sync_plans, :create, 'create a sync plan').
+      with_params('organization_id' => org_id,
+                  'name' => name,
+                  'interval' => hourly,
+                  'sync_date' => date,
+                  'enabled' => true)
+    command = %W(sync-plan create --organization-id #{org_id} --name #{name}
+                 --interval #{hourly} --enabled 1 --sync-date #{date})
+    assert_equal(0, run_cmd(command).exit_code)
+  end
+
+  it 'with custom cron' do
+    api_expects(:sync_plans, :create, 'create a sync plan').
+      with_params('organization_id' => org_id,
+                  'name' => name,
+                  'interval' => custom,
+                  'cron_expression' => cron,
+                  'sync_date' => date,
+                  'enabled' => true)
+    # end
+    command = %W(sync-plan create --organization-id #{org_id} --name #{name} --interval #{custom}
+                 --cron-expression #{cron} --enabled 1 --sync-date #{date})
+    assert_equal(0, run_cmd(command).exit_code)
+  end
+
+  it 'fails without organization-id' do
+    command = %w(sync-plan create --name #{name}
+                 --interval #{hourly} --enabled 1 --sync-date #{date})
+    refute_equal(0, run_cmd(command).exit_code)
+  end
+
+  it 'fails without name' do
+    command = %w(sync-plan create --organization-id #{org_id}
+                 --interval #{hourly} --enabled 1 --sync-date #{date})
+    refute_equal(0, run_cmd(command).err)
+  end
+
+  it 'fails without interval' do
+    command = %w(sync-plan create --organization-id #{org_id} --name #{name}
+                 --enabled 1 --sync-date #{date})
+    refute_equal(0, run_cmd(command).exit_code)
+  end
+
+  it 'fails without enabled' do
+    command = %w(sync-plan create --organization-id #{org_id} --name #{name}
+                 --interval #{hourly} --sync-date #{date})
+    refute_equal(0, run_cmd(command).exit_code)
+  end
+end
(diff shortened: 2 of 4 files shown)


lib/hammer_cli_katello/sync_plan.rb
```diff
@@ -9,6 +9,7 @@ class ListCommand < HammerCLIKatello::ListCommand
         field :sync_date, _("Start Date"), Fields::Date
         field :interval, _("Interval")
         field :enabled, _("Enabled"), Fields::Boolean
+        field :cron_expression, _("Cron Expression")
```
akofink: There's no validation in Katello for this. i.e. require interval to be 'custom cron'

lib/hammer_cli_katello/sync_plan.rb
```diff
@@ -32,9 +33,11 @@ class InfoCommand < HammerCLIKatello::InfoCommand
     class CreateCommand < HammerCLIKatello::CreateCommand
       option "--interval", "INTERVAL", _("how often synchronization should run"),
              :format => HammerCLI::Options::Normalizers::Enum.new(
-               %w(hourly daily weekly)
+               ["hourly", "daily", "weekly", "custom cron"]
```
akofink: Since we need added validation for cron_expression, I think we can also remove this Enum format completely and validate the interval in Katello. Right now, there is validation, but it's not clear what the options are from the error message (note, I removed this format to achieve the following output):

```
$ hammer sync-plan create --name foo3 --organization-id 1 --interval weeklys --sync-date 2018-09-11 --enabled 1
Could not create the sync plan:
  Interval not set correctly
```

sjha4:
@akofink : What does the label Requires katello PR you added mean? Also, I think sync-plans is not covered well in general in hammer..I will try to add some tests around that with this PR..

akofink:
@sjha4 The label means that this hammer PR requires the Katello PR I put in the description. It's mostly helpful for testing and packaging.

akofink:
You will need to update the apipie cache to test this new `cron_expression` option. Steps 3 through 5 of https://github.com/theforeman/hammer-cli-foreman/tree/master/test/data.

lib/hammer_cli_katello/sync_plan.rb
sjha4: Right..So my intent is to display the cron expression..Should we not do that?

akofink: This hammer change is correct, but I feel we should add validation in Katello for this field, as it's only valid when `interval == 'custom cron'`

lib/hammer_cli_katello/sync_plan.rb
sjha4: I will go ahead and remove the enum format.. Are you suggesting a change in validation message somewhere to prompt the hammer user of valid options?

akofink: Not only hammer users, but API users in general. Almost all of our error messages are generated by the API; this is a special case for some reason, and I don't think it needs to be.

sjha4:
@akofink : Wondering if the api cache merged 4 days ago for 3.9 would be enough for this?

akofink:
I think it's sufficient.

lib/hammer_cli_katello/sync_plan.rb
omkarkhatavkar: Can we change the name of 'custom cron' to 'custom-cron'. 
Reasons : 
1. Autocomplete is hammer shell show 'custom cron', if you see hammer shell whenever there are two names are joined using '-'  e.g. 'sync-plan', 'organization-id'
below example explains more too 


```
hammer> sync-plan create --name 'test1' --interval custom cron --organization-id 1 --enabled 1 --sync-date '2016-10-01 01:00:00'
Could not create the sync plan:
  Error: Option '--interval': Value must be one of 'hourly', 'daily', 'weekly', 'custom cron'..
  
  See: ' sync-plan create --help'.
hammer> sync-plan create --name 'test1' --interval daily --organization-id 1 --enabled 1 --sync-date '2016-10-01 01:00:00'Sync plan created.
```
2. As per above example the for daily get autocomplete no need to user to convert it to 'daily', same we can do custom-cron instread of 'custom cron'

@akofink , @sjha4  Let me know what you think ?

akofink: I don't think we should change it. Hyphens are used for subcommands and options (i.e. --organization-id), but not for values (i.e. 'Default Organization'). Values with spaces may use quotes or escape the space (i.e. `'Default Organization'` or `Default\ Organization`). This is common throughout the different hammer resources.

omkarkhatavkar: Ok,  make sense

sjha4:
@omkarkhatavkar : Ready to retest..

omkarkhatavkar:
**Scenario_1 got passed now** 

```
hammer> sync-plan update --name test --interval 'custom cron' --enabled yes --sync-date "2016-10-01 01:00:00" --organization-id 1 --cron-expression '* * 5 * *'
Sync plan updated.
hammer> sync-plan update --name test --interval 'custom cron' --enabled yes --sync-date "2016-10-01 01:00:00" --organization-id 1 --cron-expression '* * 4 * *'
Sync plan updated.

```

as @akofink said, yes the sceanario_3 (validation) is still failing.

sjha4:
@akofink, @omkarkhatavkar : Scenario 3 should give an error now that https://github.com/Katello/katello/pull/7700 is merged..

akofink:
Still erroring on scenario 3 (and yes, I tested this with https://github.com/Katello/katello/pull/7700 checked out before it was merged):

```sh
$ hammer sync-plan create --name foo4 --organization-id 1 --interval weekly --sync-date 2018-09-11 --enabled 1 --cron-expression "foo"
Sync plan created.
```

sjha4:
Maybe @omkarkhatavkar can confirm this behavior..I spun up a new hammer-devel box and pulled my changes..Works for me on a fresh setup.. :confused:

akofink:
I figured out why mine wasn't working! Once you generate an apipie cache on your Foreman instance, it uses that for all apidoc requests, even with `hammer -r` apparently (navigating to https://centos7-devel/apidoc/v2/sync_plans/create.html showed the missing cron_expression param). Regenerating the cache fixed scenario 3 for me. :)

@sjha4 You should add a test or two for this (there are no sync_plan create tests currently).

akofink:
As for scenario 2, let's file a new issue and address it in a separate PR.

sjha4:
@akofink: Added some tests..

sjha4:
Forgot about rubocop..

test/functional/sync_plan/create_test.rb
```diff
@@ -0,0 +1,61 @@
...
+require_relative 'sync_plan_helpers'
+require_relative '../organization/organization_helpers'
+
+module HammerCLIKatello
+  describe SyncPlan::CreateCommand do
+    include OrganizationHelpers
+    include SyncPlanHelpers
+
+    it 'create with organization ID,name,interval,date and enabled' do
+      api_expects(:sync_plans, :create, 'create a sync plan') do |par|
```
akofink: @sjha4 sorry, should have mentioned, there's a new syntax (`.with_params`) which makes the test failures much easier to understand. i.e. https://github.com/Katello/hammer-cli-katello/blob/49b48b3c1d431da10c254e3c59e62552cb909122/test/functional/host/extensions/create_test.rb#L26

test/functional/sync_plan/delete_test.rb
```diff
@@ -0,0 +1,53 @@
+require_relative '../test_helper'
```
akofink: You don't need to add tests here. Just for create, since that's what this PR is about.

test/functional/sync_plan/update_test.rb
```diff
@@ -0,0 +1,45 @@
+require_relative '../test_helper'
```
akofink: Same here. You're off the hook for update.

test/functional/sync_plan/create_test.rb
```diff
@@ -0,0 +1,61 @@
+require_relative '../test_helper'
+require 'hammer_cli_katello/sync_plan'
+require_relative 'sync_plan_helpers'
```
akofink: You're not using this helper in any of the create tests.

sjha4:
@akofink : Updated the tests to use the with_params..I can remove the update and delete tests if you have any concerns there..

test/functional/sync_plan/update_test.rb
```diff
@@ -0,0 +1,44 @@
...
+  include OrganizationHelpers
+  include SyncPlanHelpers
+
+  let(:org_id) { 1 }
+  let(:id) { 1 }
+  let(:name) { "sync_plan1" }
+  let(:org_name) { "org1" }
+  let(:desc) { "New Description" }
+
+  it 'with organization id and sync plan ID' do
```
akofink: This is a bit weird, but I'll accept it. :) The `--organization-id` isn't really useful with `--id`. It's only used to look up the ID using the sync plan name, but I guess it could make sense to test that organization ID *isn't* being sent to the API, which is one thing this test does.

PR #588, opened 2018-09-18
Fixes #18340 - Consistent hammer params for ordering

ofedoren:
Requires https://github.com/Katello/katello/pull/7707 to be merged first as it removes `--by` option automatically and adds two new options `--sort-by` and `--sort-order` which should be also removed according to the issue.

### DIFF
--- a/lib/hammer_cli_katello/commands.rb
+++ b/lib/hammer_cli_katello/commands.rb
@@ -43,6 +43,13 @@ module HammerCLIKatello
 
   class ListCommand < HammerCLIForeman::ListCommand
     include HammerCLIKatello::ResolverCommons
+
+    def self.build_options(builder_params = {}, &block)
+      # remove --sort-by and --sort-order in favor of the Foreman's --order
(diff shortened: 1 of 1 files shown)


lib/hammer_cli_katello/commands.rb
```diff
@@ -43,6 +43,13 @@ class SingleResourceCommand < HammerCLIForeman::SingleResourceCommand
 
   class ListCommand < HammerCLIForeman::ListCommand
     include HammerCLIKatello::ResolverCommons
+
+    def self.build_options(builder_params = {}, &block)
+      # remove --sort-by and --sort-order in favor of the Foreman's --order
+      builder_params[:without] ||= []
+      builder_params[:without] += %i[sort_by sort_order]
```
akofink: This should be `%i(sort_by sort_order)`

ofedoren:
@akofink, https://github.com/Katello/katello/pull/7707 is merged, so I guess this PR is no longer blocked.

PR #589, opened 2018-09-18
Fixes #24963: Add new export functionality to export command

ehelms:
Deprecates old export command to be called export-legacy

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,9 @@
 # TODO: remove this
 inherit_from: .rubocop_todo.yml
 
+AllCops:
+  TargetRubyVersion: 2.3
+
 MethodLength:
   Description: 'Avoid methods longer than 30 lines of code.'
   Max: 30 # default is 10
--- a/lib/hammer_cli_katello/apipie_helper.rb
+++ b/lib/hammer_cli_katello/apipie_helper.rb
@@ -1,0 +1,15 @@
+module HammerCLIKatello
+  module ApipieHelper
+    def show(resource, options = {})
+      call(:show, resource, options)
+    end
+
+    def index(resource, options = {})
+      call(:index, resource, options)['results']
+    end
+
+    def call(action, resource, options)
+      HammerCLIForeman.foreman_resource(resource).call(action, options)
+    end
+  end
+end
--- a/lib/hammer_cli_katello/content_view_purge.rb
+++ b/lib/hammer_cli_katello/content_view_purge.rb
@@ -53,7 +53,7 @@ module HammerCLIKatello
     end
 
     def execute
-      if option_count < 0
+      if option_count.negative?
         output.print_error _("Invalid value for --count option: value must be 0 or greater.")
         return HammerCLI::EX_USAGE
       end
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -1,3 +1,5 @@
+require 'fileutils'
+
 module HammerCLIKatello
   class ContentViewVersion < HammerCLIKatello::Command
     resource :content_view_versions
@@ -245,17 +247,141 @@ module HammerCLIKatello
       end
     end
 
-    class ExportCommand < HammerCLIKatello::SingleResourceCommand
+    class LegacyExportCommand < HammerCLIKatello::SingleResourceCommand
       include HammerCLIForemanTasks::Async
+      desc 'Export a content view (deprecated)'
 
       action :export
-      command_name "export"
+      command_name "export-legacy"
 
       success_message _("Content view is being exported in task %{id}.")
       failure_message _("Could not export the content view")
 
       build_options do |o|
         o.expand(:all).including(:environments, :content_views, :organizations)
+      end
+    end
+
+    class ExportCommand < HammerCLIForeman::Command
+      include HammerCLI::Messages
+      include HammerCLIKatello::LocalHelper
+      include HammerCLIKatello::ApipieHelper
+
+      PUBLISHED_REPOS_DIR = "/var/lib/pulp/published/yum/https/repos/".freeze
+
+      desc 'Export a content view version'
+
+      command_name "export"
+
+      success_message _("Content view export is available in %{directory}.")
+      failure_message _("Could not export the content view")
+
+      option "--id", "ID", _("Content View Version numeric identifier")
+      option '--export-dir', 'EXPORT_DIR', _("Directory to put content view version export into.")
+
+      validate_options do
+        option(:option_export_dir).required
+      end
+
+      build_options
+
+      class ContentViewIdParamSource
+        def initialize(command)
+          @command = command
+        end
+
+        def get_options(_defined_options, result)
+          if result['option_id'].nil? && result['option_name']
+            result['option_id'] = @command.resolver.content_view_id(result)
+          end
+          result
+        end
+      end
+
+      def option_sources
+        super << ContentViewIdParamSource.new(self)
+      end
+
+      def execute
+        cvv = show(:content_view_versions, 'id' => options['option_id'])
+        repositories = cvv['repositories'].collect do |repo|
+          show(:repositories, 'id' => repo['id'])
+        end
+
+        check_repo_download_policy(repositories)
+
+        repositories.each do |repo|
+          repo['packages'] = index(:packages, 'repository_id' => repo['id'])
+          repo['errata'] = index(:errata, 'repository_id' => repo['id'])
+        end
+
+        json = export_json(cvv, repositories)
+        create_tar(cvv, repositories, json)
+        return HammerCLI::EX_OK
+      end
+
+      def create_tar(cvv, repositories, json)
+        export_prefix = "export-#{cvv['id']}"
+        export_file = "#{export_prefix}.json"
+        export_repos_tar = "#{export_prefix}-repos.tar"
+        export_tar = "#{export_prefix}.tar"
+
+        Dir.mkdir("#{options['option_export_dir']}/#{export_prefix}")
+
+        Dir.chdir(PUBLISHED_REPOS_DIR) do
+          repositories.each do |repo|
+            repo_tar = "#{options['option_export_dir']}/#{export_prefix}/#{export_repos_tar}"
+            `tar cvfh #{repo_tar} #{repo['relative_path']}`
+          end
+        end
+
+        Dir.chdir("#{options['option_export_dir']}/#{export_prefix}") do
+          File.open(export_file, 'w') do |file|
+            file.write(JSON.pretty_generate(json))
+          end
+        end
+
+        Dir.chdir(options['option_export_dir']) do
+          `tar cf #{export_tar} #{export_prefix}`
+          FileUtils.rm_rf(export_prefix)
+        end
+      end
+
+      def check_repo_download_policy(repositories)
+        on_demand = repositories.select do |repo|
+          show(:repositories, 'id' => repo['library_instance_id'])['download_policy'] == 'on_demand'
+        end
+        return true if on_demand.empty?
+
+        on_demand_names = repositories.collect { |repo| repo['name'] }
+        raise <<~MSG
+          All exported repositories must be set to an immediate download policy and re-synced.
+          The following repositories need action:
+            #{on_demand_names.join('\n')}
+        MSG
+      end
+
+      def export_json(content_view_version, repositories)
+        json = {
+          "name" => content_view_version['content_view']['name'],
+          "major" => content_view_version['major'],
+          "minor" => content_view_version['minor']
+        }
+
+        json["repositories"] = repositories.collect do |repo|
+          {
+            "id" => repo['id'],
+            "label" => repo['label'],
+            "content_type" => repo['content_type'],
+            "backend_identifier" => repo['backend_identifier'],
+            "relative_path" => repo['relative_path'],
+            "on_disk_path" => "#{PUBLISHED_REPOS_DIR}/#{repo['relative_path']}",
+            "rpm_filenames" => repo['packages'].collect { |package| package['filename'] },
+            "errata_ids" => repo['errata'].collect { |errata| errata['errata_id'] }
+          }
+        end
+
+        json
       end
     end
 
--- a/lib/hammer_cli_katello/local_helper.rb
+++ b/lib/hammer_cli_katello/local_helper.rb
@@ -1,0 +1,9 @@
+module HammerCLIKatello
+  module LocalHelper
+    def parse_subcommand
+      return super if File.exist?('/usr/share/foreman')
+      raise "This command can only be run on the same server that Foreman is running on " \
+        "and cannot be run remotely."
+    end
+  end
+end
--- a/test/functional/apipie_helper_test.rb
+++ b/test/functional/apipie_helper_test.rb
@@ -1,0 +1,26 @@
+require File.join(File.dirname(__FILE__), '../test_helper')
+
+describe 'apipie helper' do
+  before do
+    class ApipieTestHelper
+      include HammerCLIKatello::ApipieHelper
+    end
+
+    @apipie_helper = ApipieTestHelper.new
+  end
+
+  it "calls show method apipie resource" do
+    api_expects(:repositories, :show).with_params('id' => '1').returns({})
+    assert @apipie_helper.show(:repositories, 'id' => '1')
+  end
+
+  it "calls index method apipie resource" do
+    api_expects(:repositories, :index).with_params('name' => 'foo').returns('results' => [])
+    assert @apipie_helper.index(:repositories, 'name' => 'foo')
+  end
+
+  it "call method for apipie resources works" do
+    api_expects(:repositories, :index).with_params('name' => 'foo').returns('results' => [])
+    assert @apipie_helper.call(:index, :repositories, 'name' => 'foo')
+  end
+end
--- a/test/functional/content_view/version/export_test.rb
+++ b/test/functional/content_view/version/export_test.rb
@@ -1,0 +1,87 @@
+require File.join(File.dirname(__FILE__), '../../test_helper')
+
+describe 'content-view version export' do
+  include ForemanTaskHelpers
+
+  before do
+    @cmd = %w(content-view version export)
+  end
+
+  it "performs export" do
+    params = [
+      '--id=5',
+      '--export-dir=/tmp/exports'
+    ]
+
+    ex = api_expects(:content_view_versions, :show)
+    ex.returns(
+      'id' => '5',
+      'repositories' => [{'id' => '2'}],
+      'major' => 1,
+      'minor' => 0,
+      'content_view' => {'name' => 'cv'}
+    )
+
+    ex = api_expects(:repositories, :show).with_params('id' => '2')
+    ex.returns(
+      'id' => '2',
+      'label' => 'Test_Repo',
+      'content_type' => 'yum',
+      'backend_identifier' => 'Default_Organization-Library-Test_Repo',
+      'relative_path' => 'Default_Organization/Library/Test_Repo',
+      'library_instance_id' => '1'
+    )
+
+    api_expects(:repositories, :show).with_params('id' => '1').returns(
+      'id' => '1',
+      'download_policy' => 'immediate'
+    )
+    api_expects(:packages, :index).returns('results' => [])
+    api_expects(:errata, :index).returns('results' => [])
+
+    File.expects(:exist?).with('/usr/share/foreman').returns(true)
+    Dir.expects(:chdir).with("/var/lib/pulp/published/yum/https/repos/").returns(true)
+    Dir.expects(:mkdir).with('/tmp/exports/export-5').returns(0)
+    Dir.expects(:chdir).with('/tmp/exports').returns(0)
+    Dir.expects(:chdir).with('/tmp/exports/export-5').returns(0)
+
+    result = run_cmd(@cmd + params)
+    assert_equal(0, result.exit_code)
+  end
+
+  it "fails export if any repository is set to on_demand" do
+    params = [
+      '--id=5',
+      '--export-dir=/tmp/exports'
+    ]
+
+    ex = api_expects(:content_view_versions, :show)
+    ex.returns(
+      'id' => '5',
+      'repositories' => [{'id' => '2'}],
+      'major' => 1,
+      'minor' => 0,
+      'content_view' => {'name' => 'cv'}
+    )
+
+    ex = api_expects(:repositories, :show).with_params('id' => '2')
+    ex.returns(
+      'id' => '2',
+      'label' => 'Test_Repo',
+      'content_type' => 'yum',
+      'backend_identifier' => 'Default_Organization-Library-Test_Repo',
+      'relative_path' => 'Default_Organization/Library/Test_Repo',
+      'library_instance_id' => '1'
+    )
+
+    api_expects(:repositories, :show).with_params('id' => '1').returns(
+      'id' => '1',
+      'download_policy' => 'on_demand'
+    )
+
+    File.expects(:exist?).with('/usr/share/foreman').returns(true)
+
+    result = run_cmd(@cmd + params)
+    assert_equal(70, result.exit_code)
+  end
+end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -13,3 +13,6 @@ Metrics/AbcSize:
 # Offense count: 1
 Metrics/PerceivedComplexity:
   Max: 8
+
+Style/FrozenStringLiteralComment:
+  Enabled: False
--- a/lib/hammer_cli_katello.rb
+++ b/lib/hammer_cli_katello.rb
@@ -27,6 +27,8 @@ module HammerCLIKatello
   require 'hammer_cli_katello/id_resolver'
   require 'hammer_cli_katello/capsule'
   require 'hammer_cli_katello/id_name_options_validator'
+  require 'hammer_cli_katello/local_helper'
+  require 'hammer_cli_katello/apipie_helper'
 
   # commands
   HammerCLI::MainCommand.lazy_subcommand("activation-key", _("Manipulate activation keys"),
--- a/test/functional/local_helper_test.rb
+++ b/test/functional/local_helper_test.rb
@@ -1,0 +1,30 @@
+require File.join(File.dirname(__FILE__), '../test_helper')
+
+describe 'local helper' do
+  before do
+    class SuperLocalTestHelper
+      def parse_subcommand
+        true
+      end
+    end
+
+    class LocalTestHelper < SuperLocalTestHelper
+      include HammerCLIKatello::LocalHelper
+    end
+
+    @local_helper = LocalTestHelper.new
+  end
+
+  it "does not fail if /usr/share/foreman exists" do
+    File.expects(:exist?).with('/usr/share/foreman').returns(true)
+    assert @local_helper.parse_subcommand
+  end
+
+  it "fails if /usr/share/foreman does not exist" do
+    File.expects(:exist?).with('/usr/share/foreman').returns(false)
+
+    assert_raises RuntimeError do
+      @local_helper.parse_subcommand
+    end
+  end
+end


lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +259,130 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
...
+        json = export_json(cvv, repositories)
+        create_tar(cvv, repositories, json)
+      end
+
+      def create_tar(cvv, repositories, json)
+        export_file = "export-#{cvv['id']}.json"
+        export_repos_tar = "exported-repos-#{cvv['id']}.tar"
+        export_tar = "export-#{cvv['id']}.tar"
+
+        Dir.chdir("/var/lib/pulp/published/yum/https/repos/") do
```
ekohl: Hardcoded paths always trigger me. If this shouldn't be configurable, at least make it a constant please :)

test/functional/content_view/version/export_test.rb
```diff
@@ -0,0 +1,48 @@
+require File.join(File.dirname(__FILE__), '../../test_helper')
```
ekohl: There is `__dir__` that's equal to `File.dirname(__FILE__)`. Otherwise there's `require_relative` as well.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +259,130 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
...
+      end
+
+      def export_json(content_view_version, repositories)
+        json = {
+          "cv_name": content_view_version['content_view']['name'],
+          "exported_cvv_major": content_view_version['major'],
+          "exported_cvv_minor": content_view_version['minor']
+        }
+
+        json["repositories"] = repositories.collect do |repo|
```
beav: This part looks good, but when you transform the JSON for the import, keep in mind that the new JSON is a little bit "flatter" than the old json from the proof of concept :pancakes: . See https://github.com/Katello/katello/pull/7696

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -245,11 +245,11 @@ def request_params_hosts(params)
       end
     end
 
-    class ExportCommand < HammerCLIKatello::SingleResourceCommand
+    class LegacyExportCommand < HammerCLIKatello::SingleResourceCommand
       include HammerCLIForemanTasks::Async
 
       action :export
-      command_name "export"
+      command_name "export-legacy"
```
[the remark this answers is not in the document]
ehelms: Normally true, this is a tough case where we want export to be the new workflow, but give users access to the old workflow. I could not think of a better way to do this other than renaming the old command to prevent carrying around some weird command for new functionality. Also, I still haven't figure dout how to "deprecate" a command.

lib/hammer_cli_katello/content_view_version.rb
ehelms: Can ya give me an example of what the new export JSON might should then lok like? Take a look at https://github.com/Katello/hammer-cli-katello/pull/589/files#diff-c16d36ded485c9e83859c586343e3bb6R362

test/functional/content_view/version/export_test.rb
ehelms: That a good idea, I just took this from other test files, so gonna err on the side of consistency.

beav:
@ehelms answering here since the other reply got hidden by github. Your export json looks good, my comment was for when you do the import side. When you parse it to call `/publish` based on the `import.sh` proof of concept, the old style had something like `{ repos_units: { repos: [ stuff ] } }` but the new way is just `{repos_units: [ stuff ] }`.

lib/hammer_cli_katello/content_view_version.rb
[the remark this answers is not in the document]
akofink: This is the same action as the old export command. Is it not possible to make them the same command, differentiating behavior with some flag/option? This would be the first hammer command to duplicate an existing API endpoint, a red flag in my opinion.

akofink:
This command, unlike the other hammer commands, must be run on the Katello host. I think we should check that Katello is indeed installed, rather than this somewhat cryptic message:

```
$ hammer content-view version export --id 20 --export-dir /tmp/
Could not export the content view:
  Error: No such file or directory @ dir_chdir - /var/lib/pulp/published/yum/https/repos/
```

lib/hammer_cli_katello/content_view_version.rb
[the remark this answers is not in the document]
akofink: Since the old export is allowed from a remote hammer machine, perhaps the flag could be something like `--remote` for the old command or `--local` for the new command (or something similar with more detail; naming is hard).

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +259,131 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
...
+
+        json = export_json(cvv, repositories)
+        create_tar(cvv, repositories, json)
+      end
+
+      def create_tar(cvv, repositories, json)
+        export_file = "export-#{cvv['id']}.json"
+        export_repos_tar = "exported-repos-#{cvv['id']}.tar"
+        export_tar = "export-#{cvv['id']}.tar"
+        published_repo_dir = "/var/lib/pulp/published/yum/https/repos/"
```
akofink: Let's make this a constant. You're using it as a literal in 3 places.

lib/hammer_cli_katello/content_view_version.rb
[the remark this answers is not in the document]
ehelms: I think you are thinking in the right idea, but conceptually it puts us in a place we don't want to be. We aren't looking to support remote and local and have them two completely different processes. We want to move users to this new way of doing export and import as the only way we support doing it. With that in mind, any other thoughts how we could differentiate?

lib/hammer_cli_katello/content_view_version.rb
ehelms: Given this is only used within the context of this function, shouldn't I prefer to keep it scoped?

lib/hammer_cli_katello/content_view_version.rb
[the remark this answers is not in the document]
akofink: > We aren't looking to support remote and local and have them two completely different processes

Great, it sounds like we want to support remote and local and have them be the same process (i.e. with `hammer content-view version export`). This PR currently fails to export when running the hammer command from a different box. Were you planning to support (or handle) this? There need not be a flag for something we can determine programmatically.

lib/hammer_cli_katello/content_view_version.rb
akofink: Not sure I understand your concern?

lib/hammer_cli_katello/content_view_version.rb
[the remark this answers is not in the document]
ehelms: Let me clarify. We want to support only the new method for export and import. That method requires local hammer only. There is no support for running hammer remotely to perform this new export and import.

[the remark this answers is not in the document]
akofink: > That method requires local hammer only

The first hammer command in hammer-cli-katello of its kind afaik. Perhaps this remote CLI tool isn't the best place to put local-only export (i.e. hammer-cli-foreman-admin is local only).

Again, you must at least handle the remote case because hammer-cli-katello is a remote CLI tool. All I'm suggesting is that you handle it by invoking the old export functionality, which is the first step of your new functionality anyway (just the API call).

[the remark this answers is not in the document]
ehelms: The old export is entirely different than the new export in terms of how its designed to work. We are deprecating that API to enforce this as well. This is Katello functionality however, so spinning up an entire hammer project just to handle this seems like overkill. There is no clean way to provide "remote" support for this functionality. So I think we either have this special snowflake or we don't provide users a simple, standard orchestrated way to export/import content views. The alternative is to push all of this functionality to Ansible, however, we have to solve a lot of deliverable questions if we do that.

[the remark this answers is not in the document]
akofink: > We are deprecating that API to enforce this as well

Don't you rely on that API for the new export command? Where's the Katello PR to deprecate the API?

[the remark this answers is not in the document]
ehelms: We do not use that export API at all for new export/import. On the to do list to deprecate it.

ehelms:
The related import command PR -- https://github.com/Katello/hammer-cli-katello/pull/592

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +259,131 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
       end
     end
 
+    class ExportCommand < HammerCLIKatello::SingleResourceCommand
+      include HammerCLIForemanTasks::Async
+
+      action :export
```
akofink: > We do not use that export API at all for new export/import. On the to do list to deprecate it.

This is why I'm confused, I guess.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +259,131 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
...
+      def export_json(content_view_version, repositories)
+        json = {
+          "name" => content_view_version['content_view']['name'],
+          "major" => content_view_version['major'],
+          "minor" => content_view_version['minor']
+        }
+
+        json["repositories"] = repositories.collect do |repo|
+          {
+            "repo_id" => repo['id'],
```
akofink: Why the `repo_` prefix? IMO these fields are all repo fields if they're in an array pointed to by the "repositories" key. We don't do this in Katello; what's the motivation here?

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +259,131 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
...
+
+      def packages(id)
+        HammerCLIForeman.foreman_resource(:packages).call(:index, 'repository_id' => id)['results']
+      end
+
+      def errata(id)
+        HammerCLIForeman.foreman_resource(:errata).call(:index, 'repository_id' => id)['results']
+      end
+
+      def export_json(content_view_version, repositories)
```
akofink: This sure does seem like something Katello should be doing, not hammer. Explain again why we're removing the API endpoint from Katello and opting to use hammer to cobble together JSON output? I wasn't involved in planning this feature, but this doesn't seem to be the right way to implement it IMO.

chris1984:
Also a general usage question, here is my versions below:

~~~
hammer> content-view version list 
---|-------------------------------|---------|-----------------------
ID | NAME                          | VERSION | LIFECYCLE ENVIRONMENTS
---|-------------------------------|---------|-----------------------
3  | Zoo 4.1                       | 4.1     | Library               
2  | Zoo 1.0                       | 1.0     |                       
1  | Default Organization View 1.0 | 1.0     | Library               
---|-------------------------------|---------|-----------------------
```

I see we have the option to pass in id and version, should we remove version since we have to pass in the id from above, or can we pass in version instead of id?

File: lib/hammer_cli_katello/content_view_version.rb (hunk 6 above)
Comment (ehelms) [reply lib/hammer_cli_katello/content_view_version.rb]: I'll be honest, I just assumed this was boilerplate. If not using that API, does this line of code go away?

File: lib/hammer_cli_katello/content_view_version.rb (hunk 7 above)
Comment (ehelms) [reply lib/hammer_cli_katello/content_view_version.rb]: This was following the original design by @beav . I do agree I think this can be simplified. Will let him weigh in

File: lib/hammer_cli_katello/content_view_version.rb (hunk 8 above)
Comment (ehelms) [reply lib/hammer_cli_katello/content_view_version.rb]: We could push this into an API in Katello. That API would be incomplete with respect to generating an export. An export needs content + the export JSON. The idea was this centers users to *have* to follow this workflow to get an export. I'll let @beav weigh in more

**Comment by ehelms** [comment]
@chris1984 Updated and dropped those parameters that aren't used

File: lib/hammer_cli_katello/content_view_version.rb (hunk 6 above)
Comment (akofink) [reply lib/hammer_cli_katello/content_view_version.rb]: You definitely don't need it if you're not using hammer's API resource mechanism, though you may need to override other methods (execute may be sufficient, which you're already overriding), though the ID resolution code uses this as well. Ultimately, you should remove it and see what breaks - it should be code that hammer is running that you're not using for this snowflake command anyway and so should arguably be stubbed.

**Comment by akofink** [comment]
@ehelms @beav I'd encourage you to consider other projects in which to implement this feature, as it's kind of shoe horned into hammer-cli-katello, not using any of the provided hammer API features and breaking the remote hammer paradigm. What about foreman-maintain, a tool that's installed by default in production and is expected to be used by an admin locally on the Foreman system? I hope you can understand my perspective and reasons for push back here. :hammer:

**Comment by ehelms** [comment]
I get your concerns. I will rule out foreman-maintain because I think shoe-horning this into that is worse than using hammer. I see foreman-maintain as the tool that helps you perform maintenance on your server. While this is providing data functionality. I'll be honest, if don't add export and import support to hammer then we are going to leave as an exercise to script a set of steps. That feels worse than putting this in hammer. One reason we went down this route, the community requested this be in hammer -- https://community.theforeman.org/t/content-view-import-export/10608/3

**Comment by akofink** [comment]
@ehelms One of the things the community expects by using hammer is the [ability to import/export from different hosts](https://community.theforeman.org/t/content-view-import-export/10608/6). There wasn't much discussion about that on the post, but IMO I agree that this would be a good reason to consider hammer, and local-only is one of my main concerns for this current implementation in hammer. What's the pushback for supporting remote import/export? Did that come from the community?

**Comment by ehelms** [comment]
To support different hosts, we would need to add SSH supporting code and workflow to hammer. To us, that felt ugly and something we should avoid.

**Comment by akofink** [comment]
Hammer and Foreman already support http, and http supports file transfer. Why the need for ssh?

**Comment by ehelms** [comment]
If you dig into the code, you'l see that export has two pieces:

 1) Build the JSON file that describes the export
 2) Extract and tar up all of the physical RPMs for each repository in the export

The second step is the part that requires physical access to the host to perform.

**Comment by akofink** [comment]
Right, I'm suggesting Katello handle the second step and send the tar over http. I don't currently see why that's not possible.

**Comment by ehelms** [comment]
This was considered, and @beav who has strong opinions and experiences with the existing method can give a better explanation of why we chose the route we did.

**Comment by beav** [comment]
> Right, I'm suggesting Katello handle the second step and send the tar over http. I don't currently see why that's not possible.

If Katello creates the tar, it will require an additional intermediate location to store it before the user copies it off the filesystem. These tars can be >500GB, and we've received user feedback that the import/export process requires too much temporary disk space. It will also create a task that takes hours and hours to complete.

If the user is given the command to create the tar, they can create it directly where they want it (USB key, remote mount, etc) and avoid this intermediate step. They'll still need to incur a lot of disk IO but it only needs to be done once and not twice.

The file transfer is also tricky since most environments have strict network restrictions. For example, most users will ssh to their Katello instance via a [jump server](https://en.wikipedia.org/wiki/Jump_server). It's extremely uncommon for a jump server to have enough disk space to hold a content export.

**Comment by akofink** [comment]
Thank you @beav for this insight from our users. I'm convinced of the local-only aspect of this import/export feature. Though, I'm still unconvinced that hammer-cli-katello is the best place for such a local-only, admin-only function. I also agree that putting it in foreman-maintain would be inappropriate.

I encourage you to try to think of some creative solutions. For example, the `hammer admin ...` sub-commands from hammer-cli-foreman-admin are all local-only, admin-only commands. I don't think hammer-cli-foreman-admin is the appropriate repo for this functionality, as it's specific to Katello. Perhaps we can add `hammer admin content-view-version-export`, defined within this repository (hammer-cli-katello), forming a new paradigm for these kinds of administrative data-related tasks in Katello. I'm quite certain this is possible without interfering with the hammer-cli-foreman-admin subcommands (so long as the subcommands are unique between the repos, of course). I'm all ears for other solutions. :ear_of_rice:

**Comment by ehelms** [comment]
I agree that we should do something to handle the local-only nature of this. I feel `hammer admin` is a bit odd of scoping for users to go find this functionality.

**Comment by ehelms** [comment]
For some clarity, I tested hammer-cli-admin and it does not do any scoping of whether ran locally or remotely. The plugin simply attempts to perform it's actions and debug logs if it cannot find the components its looking for. So as far as I can tell, we have no model for this local only currently.

**Comment by ekohl** [comment]
IMHO if it's a local only action then hammer I wonder if hammer is even the right tool. Hammer has the expectation that it can be installed on a remote host and I'm not sure we should break that. I'd be interested to see what @mbacovsky thinks about it.

**Comment by beav** [comment]
There was another discussion about whether or not this should be in hammer in https://community.theforeman.org/t/content-view-import-export/10608/3, pasting link here just to cross-ref the two threads

**Comment by ekohl** [comment]
I think the crucial assumption there was that hammer is a remote tool. Thus it was implied that this should be a remote action. To quote Bernard:
> it would also be possible to export / import a CV on a different host.

**Comment by ekohl** [comment]
I don't disagree with that things can be large and we don't have the proper infra in place to support an API, but in the RFC people suggested hammer with the assumption that it would be possible to do remote, as I thought all hammer functionality already was. Now it looks like we made incorrect assumptions (something that has never happened before), we can no longer just continue with the choices we made. I'm not saying that it's the wrong choice per se, but with new information it's good to evaluate your choices.

Another option can be to introduce a new command in the same package that exposes local commands. Not saying that's the best idea, but it would have the advantage that it's obvious to users what's local and what's remote.

This may all feel like a lot of effort, but we've learned that making a wrong architectural choice can haunt you for a long time so let's try to avoid it if we can.

**Comment by beav** [comment]
> I don't disagree with that things can be large and we don't have the proper infra in place to support an API, but in the RFC people suggested hammer with the assumption that it would be possible to do remote, as I thought all hammer functionality already was. Now it looks like we made incorrect assumptions (something that has never happened before), we can no longer just continue with the choices we made. I'm not saying that it's the wrong choice per se, but with new information it's good to evaluate your choices.
> 
> Another option can be to introduce a new command in the same package that exposes local commands. Not saying that's the best idea, but it would have the advantage that it's obvious to users what's local and what's remote.

Do you mean like under https://github.com/theforeman/hammer-cli-foreman-admin ? I wouldn't have any issue with the command living there.

**Comment by ekohl** [comment]
> Do you mean like under https://github.com/theforeman/hammer-cli-foreman-admin ? I wouldn't have any issue with the command living there.

It looks like it's already aware of "katello" components (qpid, candlepin, pulp) and works on the local server so it appears to be a better fit. However, I'm not an expert on hammer so I'll let others weigh in.

**Comment by ehelms** [comment]
I am just not sure the theme of hammer-cli-foreman-admin fits this functionality. As I understand it, that is for "admin" tasks on your installation itself. Now, you could argue that all of that functionality should simply go into foreman-maintain now that we have it. But that is getting away from this.

Using Katello functionality, we want to provide users with an import/export experience that matches the evolving requests of users in this area. This is specific to Katello, and this is unfortunately local only. 

From my looking at information on hammer-cli, and the Foreman documentation, we don't mention the 'remote' nature of our CLI. We simply talk about it being the command line interface for performing functions. I get the general idea that most CLIs can be run remotely I am just failing to see what this hurts.

I feel like our options have come down to:

 a) Go this route and accept these special local only options and give users a fluid experience
 b) Drop hammer support all together, document how to perform this using bash or python or some example language and give them maybe less great of an experience
 c) Create an entirely new hammer git repo, packaging, testing, etc. just to hold import/export
 d) Perform this all in Ansible, missing this release for providing users an orchestrated method outside of bash commands
 e) Other

We have about a week left in this development cycle.

**Comment by akofink** [comment]
Choices b - d are really not reasonable (I see what you're doing @ehelms ;). I think we can move forward with this remote/local distinction in hammer, but we should design for it. Perhaps a mixin that lives in hammer-cli-foreman to enforce local-only commands? The mixin would be responsible for determining if the hammer command is invoked locally, and if not, handling/presenting an appropriate error to the user. It should also (optionally?) 'turn off' all the built in apipie logic, ID resolution, and option builders for that command.

**Comment by beav** [comment]
one small comment, doing A doesn't preclude us from doing another option as well. I was thinking we'd do A and still follow up with D. We could also do A as-is and then create a redmine issue for the mixin perhaps.

**Comment by ehelms** [comment]
Just trying to arrive at an answer based on available choices for our users @akofink :) And I like your suggestions from a coding, re-use, providing a pattern for the entire ecosystem point of view. What I am having a hard time reconciling is how to do all of that in the limited time frame of the impending release. That leaves me with, I think:

 a) push hard to knock it out, get it accepted in a week
 b) try to get just the Foreman changes in in the next week, and do these katello changes in 3.10

The caveat being, 3.10 is likely during the 1.20 time frame, so if I cannot get changes into hammer-cli-foreman in the next week, this path is a full 3 month release cycle out.

**Comment by akofink** [comment]
We could also write it all in this repo and migrate the reusable bits out to hammer-cli-foreman in a second PR that doesn't have to land in 3.9 or 3.10. This PR didn't check/handle the remote case when I last tested it, so adding a generic mixin that does so seems reasonable. What do you think?

**Comment by ehelms** [comment]
@akofink Did you have any thoughts for how to robustly determine remote vs. local? The best I had was checking for the existence of `katello` RPM

**Comment by akofink** [comment]
Yes, that's basically what I had in mind as well, though we should take care to be as plugin-generic as possible, since we want it to be added to hammer-cli-foreman eventually.

**Comment by ehelms** [comment]
I've added a little helper. Not sure if it'sthe best design but its a starter. What do you think @akofink

File: lib/hammer_cli_katello/local_helper.rb [hunk 9]
```diff
@@ -0,0 +1,10 @@
+module HammerCLIKatello
+  module LocalHelper
+
+    def local_only
+      return true if system("rpm -q foreman", :out => File::NULL)
+      fail "Export can only be run on the server and not remotely."
```
beav: I would change to

"This command can only be run on the same server that Foreman is running on, and cannot be run remotely".

ekohl:
Let's look at it from a users perspective. If we code it now in hammer-cli-katello and later decide to 'do it right', will that impact the user or is there a (proper) way to do it all transparent to the user?

ehelms:
If the 'do it right' is related to having code in hammer-cli-foreman to check for local only commands, I think this will be seamless to the user.

lib/hammer_cli_katello/local_helper.rb
```diff
@@ -0,0 +1,12 @@
+module HammerCLIKatello
+  module LocalHelper
+    def local_only
+      return true if system("rpm -q foreman", :out => File::NULL)
```
ehelms: @akofink In the tests, I can't figure out how to either mock `system` or mock the overall call to `local_only`. I have tried many a thing with no luck.

akofink: It'd be easy to stub if the system call was in its own (private) method, and it'd be more descriptive imo.

ekohl: That method could also use better method that'd be compatible with non-RPM systems. Perhaps check for the existence of `/usr/share/foreman` for example.

beav:
@ehelms small fix:

```
# hammer content-view version export --export-dir /tmp/tmp.KoYHgSvTbT --id 2
Could not export the content view:
  Error: File exists @ dir_s_mkdir - /tmp/tmp.KoYHgSvTbT/export-2
```

A better error may be good here, to tell people to either delete the location or pick a new one.

ehelms:
@beav How did you get into that state? if I re-run the command over and over I get not errors

beav:
@ehelms I used the older patch with the missing `=`, you should be able to repro like that to induce an error. I suppose it could also happen if you ran out of disk space when creating the tar.

ehelms:
Ohh I see how you could get in to that place. Do think we should warn and make the user clean it up or have the command clean up prior to running?

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +261,137 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
       end
     end
 
+    class ExportCommand < HammerCLIKatello::SingleResourceCommand
```
akofink: This should probably be:

```rb
class ExportCommand < HammerCLI::AbstractCommand
  include HammerCLI::Messages
  ...
end
```

Using any command classes that have the `HammerCLI::Apipie::Command` ancestor will give you apipie-knowledgeable commands, and this command doesn't associate with an apipie controller action definition.

ehelms: Making that change fails earlier:

```
TypeError (no implicit conversion from nil to integer):
    /root/hammer-cli-katello/lib/hammer_cli_katello/content_view_version.rb:303:in `insert'
    /root/hammer-cli-katello/lib/hammer_cli_katello/content_view_version.rb:303:in `option_sources'
```

Any idea what is causing that or how to fix it based on the class change?

beav:
The export worked OK for me and produced the expected results, but I did not do a full end2end test with the import yet.

lib/hammer_cli_katello/local_helper.rb
```diff
@@ -0,0 +1,12 @@
+module HammerCLIKatello
+  module LocalHelper
+    def parse_subcommand
+      return super if File.exist?('/usr/share/foreman')
+      message = [
+        "This command can only be run on the same server that Foreman is running on,",
+        "and cannot be run remotely."
+      ]
+      raise message.join(' ')
```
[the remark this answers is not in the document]
akofink: It does? Backslash is your friend.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +261,134 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
...
+        HammerCLIForeman.foreman_resource(:repositories).call(:show, 'id' => id)
+      end
+
+      def packages(id)
+        HammerCLIForeman.foreman_resource(:packages).call(:index, 'repository_id' => id)['results']
+      end
+
+      def errata(id)
+        HammerCLIForeman.foreman_resource(:errata).call(:index, 'repository_id' => id)['results']
+      end
```
akofink: Looks like we should define another helper. Again, I don't believe we should be constructing the json in hammer (that's why no other hammer commands require the raw data like this). I'd *much* prefer Katello be able to return the JSON structure you're after, whether via API flag in `content_view_versions#show` or a new API endpoint. But, if we really want it in hammer, let's at least put it in a helper:

i.e.

```rb
      def show(resource, options)
        HammerCLIForeman.foreman_resource(resource).call(:show, options)
      end

      def index(resource, options)
        HammerCLIForeman.foreman_resource(resource).call(:index, options)['results']
      end
```

lib/hammer_cli_katello/content_view_version.rb
akofink: I'm suggesting you put it in `ExportCommand`:

```rb
class ExportCommand
  PUBLISHED_REPO_DIR = /var/lib/pulp/published/yum/https/repos/
  ...
end
```

In your tests, you can access the constant with `HammerCLIKatello::ContentViewVersion::ExportCommand::REPO_DIR`

lib/hammer_cli_katello/content_view_version.rb
ehelms: Sorry, I don't follow, I have methods like this already. Renaming them to index or show might make them more generic and less readable given multiple show calls are made?

lib/hammer_cli_katello/local_helper.rb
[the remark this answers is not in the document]
ehelms: Looks like backslash is pretending to be your friend with all that whitespace it adds:

```
  Error: This command can only be run on the same server that Foreman is running on,         and cannot be run remotely.
```

lib/hammer_cli_katello/content_view_version.rb
akofink: I actually think these two methods make the calls *more* readable. i.e.

```rb
# this
show(:content_view_versions, 'id' => options['option_id'])

# instead of this
HammerCLIForeman.foreman_resource(:content_view_versions).call(:show, 'id' => options['option_id'])
```

You don't think it's more readable that way? Import also does some of this type of thing, so I think having generic methods in a helper outside this class would be an improvement. In addition, you can DRY up the index logic of fetching the `['results']`, and for import, you can support ID resolution from name + other required params for show queries, which would address [this comment](https://github.com/Katello/hammer-cli-katello/pull/592/files#r222803347).

lib/hammer_cli_katello/local_helper.rb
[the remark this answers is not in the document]
ehelms: Swapping to wrapping each string part in quotes seems to have worked, I switched to + and rubocop actually warned me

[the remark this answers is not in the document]
akofink: Ah sorry, I didn't show you how to do it correctly:

```rb
raise "This command can only be run on the same server that Foreman is running on, "\
  "and cannot be run remotely."
```

IMO rubocop should never make you think to 'program around' it by using different types or performing variable manipulation. If you want a string, use a string.

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -53,7 +53,7 @@ def option_sources
     end
 
     def execute
-      if option_count < 0
+      if option_count.negative?
```
ehelms: I had to change this one line due to the rubocop changes.

.rubocop.yml
```diff
@@ -1,6 +1,9 @@
 # TODO: remove this
 inherit_from: .rubocop_todo.yml
 
+AllCops:
+  TargetRubyVersion: 2.3
```
ehelms: This was set to the default 2.0 which won't let some of the code in here through. I set this to the lowest tested Ruby version.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +261,145 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
...
+        check_repo_download_policy(repositories)
+
+        repositories.each do |repo|
+          repo['packages'] = packages(repo['id'])
+          repo['errata'] = errata(repo['id'])
+        end
+
+        json = export_json(cvv, repositories)
+        create_tar(cvv, repositories, json)
+        return 0
```
ehelms: @akofink This seems kind augly, but couldnt find another work around

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +261,145 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
...
+          repo['download_policy'] == 'on_demand'
+        end
+        return true if on_demand.empty?
+
+        on_demand_names = repositories.collect { |repo| repo['name'] }
+        raise <<~MSG
+          All exported repositories must be set to an immediate download policy and re-synced.
+          The following repositories need action:
+            #{on_demand_names.join('\n')}
+        MSG
```
ehelms: @akofink @beav Here is the on_demand check. I used this style for the message to format the output.

lib/hammer_cli_katello/content_view_version.rb
ehelms: Honestly then, I need some help how you would structure this helper. A helper per resource type? A single helper with all of these methods put into it? is it then a specific import/export helper file? or somehow made generic?

lib/hammer_cli_katello/content_view_version.rb
akofink: I think you want `return HammerCLI::EX_OK`, and that makes it look better :)

lib/hammer_cli_katello/content_view_version.rb
akofink: I'm envisioning something very generic, one method per standard REST action, and one action-generic method (i.e. for [repositories#sync](https://github.com/Katello/hammer-cli-katello/pull/592/files#diff-c16d36ded485c9e83859c586343e3bb6R459) and others):

```rb
module ApipieHelper # or whatever name you think appropriate
  def show(resource, options)
    call(:show, resource, options)
  end

  def index(resource, options)
    call(:index, resource, options)['results']
  end

  def call(action, resource, options)
    HammerCLIForeman.foreman_resource(resource).call(action, options)['results']
  end
end
```

I think such a helper could also be moved over to hammer-cli-foreman with the `LocalHelper`.

lib/hammer_cli_katello/content_view_version.rb
akofink: The way the code is written now, the additional step of re-publishing the content view is required. If we check the library repo's sync policy, that seems to be enough to have the packages found on the filesystem.

```rb
HammerCLIForeman.foreman_resource(:repositories).call(:show, 'id' => repo['library_instance_id'])['download_policy'] == 'on_demand'
```

You might think you'd need to use the library repo's `relative_path` to find the rpm on the filesystem, but it seems pulp only downloads an rpm once, so the symlink points to the same place as it's library repo's symlink.

ehelms: Good catch! And maybe this is cheating, but I see  other value as well -- https://github.com/Katello/katello/pull/7737

lib/hammer_cli_katello/apipie_helper.rb
```diff
@@ -0,0 +1,15 @@
+module HammerCLIKatello
+  module ApipieHelper # or whatever name you think appropriate
```
akofink: Uhh... LOL

ehelms: Fixed

akofink:
Oh, @ehelms you also need to use `desc` to populate the help text:

```
$ hammer content-view version -h
Usage:
    hammer content-view version [OPTIONS] SUBCOMMAND [ARG] ...

Parameters:
 SUBCOMMAND                    Subcommand
 [ARG] ...                     Subcommand arguments

Subcommands:
 delete                        Remove content view version
 export                         
 export-legacy                 Export a content view version
 incremental-update            Perform an Incremental Update on one or more Content View Versions
 info                          Show content view version
 list                          List content view versions
 promote                       Promote a content view version
 republish-repositories        Forces a republish of the version's repositories' metadata

Options:
```

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +261,142 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
...
+      end
+
+      def export_json(content_view_version, repositories)
+        json = {
+          "name" => content_view_version['content_view']['name'],
+          "major" => content_view_version['major'],
+          "minor" => content_view_version['minor']
+        }
```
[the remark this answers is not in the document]
ehelms: I get that this makes it appear simpler, but I think that collect in the middle will get overlooked and confuse people. I am choosing to go with some explicitness here to help readability.

test/functional/apipie_helper_test.rb
```diff
@@ -0,0 +1,21 @@
...
+describe 'apipie helper' do
+  before do
+    class ApipieTestHelper
+      include HammerCLIKatello::ApipieHelper
+    end
+
+    @apipie_helper = ApipieTestHelper.new
+  end
+
+  it "calls show method apipie resource" do
```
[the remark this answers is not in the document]
ehelms: If I don't test the show method the coverage will drop. Is that acceptable? If so, I'll just test the call method.

test/functional/content_view/version/export_test.rb
```diff
@@ -0,0 +1,92 @@
...
+    @cmd = %w(content-view version export)
+  end
+
+  it "performs export" do
+    params = [
+      '--id=5',
+      '--export-dir=/tmp/exports'
+    ]
+
+    ex = api_expects(:content_view_versions, :show)
```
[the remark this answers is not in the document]
ehelms: I can, but I don't really need to test that. I trust the the API methods will do the right things.

test/functional/content_view/version/export_test.rb
```diff
@@ -0,0 +1,92 @@
...
+    api_expects(:errata, :index).returns('results' => [])
+
+    File.expects(:exist?).with('/usr/share/foreman').returns(true)
+    Dir.expects(:chdir).with("/var/lib/pulp/published/yum/https/repos/").returns(true)
+    Dir.expects(:mkdir).with('/tmp/exports/export-5').returns(0)
+    Dir.expects(:chdir).with('/tmp/exports').returns(0)
+    Dir.expects(:chdir).with('/tmp/exports/export-5').returns(0)
+
+    result = run_cmd(@cmd + params)
+    assert_equal(0, result.exit_code)
```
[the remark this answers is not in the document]
ehelms: This seems to be the convention in use so I went with that.

lib/hammer_cli_katello/content_view_version.rb
ehelms: I'm OK with explicitness in the tests given this is used only in this local method. If you require it changed, I will.

test/functional/apipie_helper_test.rb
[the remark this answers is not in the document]
akofink: > Then you can test this method calls the call method with the appropriate args.

I'm not suggesting you remove this test, just change it and test the `call` method as well.

test/functional/content_view/version/export_test.rb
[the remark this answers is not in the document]
akofink: In my experience, it can help to figure out failures when they happen.

lib/hammer_cli_katello/content_view_version.rb
ehelms: In the end, it did not since we needed information from both.

lib/hammer_cli_katello/content_view_version.rb
akofink: > this is used only in this local method

But it's also used in [`export_json`](https://github.com/Katello/hammer-cli-katello/pull/589/files#diff-c16d36ded485c9e83859c586343e3bb6R377) (as I said, you're using it in 2 places, not including tests). Additionally, it's more explicit IMO to use a named variable/constant than a literal. It's not blocking, but I just don't understand your pushback (I did ask for clarification but haven't received a response).

ehelms:
I think I have resolved or commented back every comment up to this point.

test/functional/apipie_helper_test.rb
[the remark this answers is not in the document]
akofink: Not quite, but I'll accept it. Now that you're testing `call`, you don't need to use `api_expects` to test `show` and `index`, you can just `@apipie_helper.should_receive(:call).with(params)`

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -259,6 +262,129 @@ class ExportCommand < HammerCLIKatello::SingleResourceCommand
...
+
+        json["repositories"] = repositories.collect do |repo|
+          {
+            "id" => repo['id'],
+            "label" => repo['label'],
+            "content_type" => repo['content_type'],
+            "backend_identifier" => repo['backend_identifier'],
+            "relative_path" => repo['relative_path'],
+            "on_disk_path" => "#{PUBLISHED_REPOS_DIR}/#{repo['relative_path']}",
```
akofink: s/PUBLISHED_REPOS_DIR/PUBLISHED_REPO_DIR/

test/functional/apipie_helper_test.rb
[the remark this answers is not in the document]
ehelms: Yea, just feels weird to stub out a method that can actually be called vs. stubbing out the expectation overall in this particular case. This makes me feel safer :)

ehelms:
I think this is ready for another round o freview.

ehelms:
Thanks for the reviews @akofink !

PR #594, opened 2018-10-09
Refs #24963 - pull all packages and handle multiple repos

beav:
Previously, only the first 20 package names in a repo would be fetched
for the export json. This resulted in partial CV version created on
import.

Additionally, the `tar` command was overwriting the old repo tarball
for each new repo.

This patch sends `:full_result => true` when the full results are
needed. It also constructs a full `tar` command that includes all repo
directories.


beav:
@ehelms this is the fix for the two export bugs that @chris1984 found. I created it as a "refs" since it's closely tied to the original issue.

PR #599, opened 2018-10-17
[WIP] fixes #24082 - content view purge skips versions in composites

cfouant:
To test:
1) Create a content view with multiple versions (at least 4)
2) Create a composite view and include a version from the original content view (not latest)
3) Perform hammer content-view purge on the original content view

Prior to this fix, the content-view purge would fail on the version that was included in the composite view, and hammer purge would skip all other version deletions. Now, all unused content-view versions will be deleted.


lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -74,32 +68,20 @@ def execute
...
-    private def content_view
-      @content_view ||= resource_content_views.call(:show, 'id' => options['option_id'])
-    end
-
-    private def old_unused_versions
-      return @old_unused_versions if @old_unused_versions
-
-      all_versions = content_view['versions'].sort_by do |version|
-        [version[:major], version[:minor]]
-      end
```
akofink: We need to delete the oldest versions. We can either have the API return the versions in order by version or keep this sort. Right now if you purge a CV, it removes the most recent versions, which is not what we want.

cfouant: Good call, forgot to consider that! I've updated with the appropriate sorting, and added destroy method in apipie_helper and updated tests as well. Thanks!

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -59,6 +52,7 @@ def execute
       end
 
       # Check if there is something to do
+      old_unused_versions = purge_list
```
akofink: I'm not sure I understand why you rewrote the method instead of altering the existing method. The old method cached the result in an instance variable for reuse in subsequent calls, so a variable wasn't necessary here. :man_shrugging:

cfouant: I guess I preferred to make one resource call rather than 2. The existing method required a call to both `content_view` and `content_view_version` resources.

cfouant: I guess I preferred to make one resource call rather than 2. The existing method required a call to both content_view and content_view_version resources, which was unnecessary.

akofink: Yeah, you'd still need to alter the API call, but there is some logic that you want to keep (like sorting by version, which you had to re-implement in your method). So I'm wondering why you chose to rewrite it instead of altering it. When I submit a code change, I generally try to modify the existing code as little as possible (make the diff as small and easy to understand as possible, keep things that are already working correctly, etc.). This helps people in the future when they're looking at this commit to figure out what changed. I'd prefer you modify the original method rather than replace it for that reason.

cfouant: Honestly, I chose to rewrite it because it seemed messy. I didn't see any reason to keep an instance variable, as we'd want to refresh the information any time this method is called. I've made the requested changes. Thanks for explaining.

cfouant: Yup, that is why I thanked you for explaining and made the changes 👍

cfouant:
@akofink - nice catch - let me know what you think!

PR #600, opened 2018-10-22
Fixes #25272 - add support for composite CV import/export

beav:
Previously, CV import/export would treat composites as "regular"
content views.  This resulted in each repo of the composite being
exported as if it was part of a regular content view.

These exports were unable to be imported since the repos existed on
the components and not the composite.

This commit exports composite content views with information needed to
reconstitute them on the importing side. Each component CV name and
version is written to the JSON, and upon import, hammer searches for
the matching name and version on the importing side. If all of the
info lines up, the components are associated with the CV and a publish
occurs.

### DIFF
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -265,6 +265,7 @@ module HammerCLIKatello
     class ExportCommand < HammerCLIForeman::Command
       include HammerCLIKatello::LocalHelper
       include HammerCLIKatello::ApipieHelper
+      include HammerCLIKatello::CVImportExportHelper
 
       PUBLISHED_REPOS_DIR = "/var/lib/pulp/published/yum/https/repos/".freeze
 
@@ -287,18 +288,29 @@ module HammerCLIKatello
 
       def execute
         cvv = show(:content_view_versions, 'id' => options['option_id'])
-        repositories = cvv['repositories'].collect do |repo|
-          show(:repositories, 'id' => repo['id'], :full_result => true)
+        cv = show(:content_views, 'id' => cvv['content_view_id'])
+
+        composite = cv["composite"]
+        export_json_options = { :cvv => cvv }
+
+        if composite
+          components = cv['components']
+          export_json_options[:component_cvvs] = components.collect do |component|
+            component['name']
+          end
+          export_json_options[:repositories] = []
+        else
+          repositories = fetch_cvv_repositories(cvv)
+          check_repo_download_policy(repositories)
+
+          repositories.each do |repo|
+            repo['packages'] = index(:packages, 'repository_id' => repo['id'], :full_result => true)
+            repo['errata'] = index(:errata, 'repository_id' => repo['id'], :full_result => true)
+          end
+          export_json_options[:repositories] = repositories
         end
 
-        check_repo_download_policy(repositories)
-
-        repositories.each do |repo|
-          repo['packages'] = index(:packages, 'repository_id' => repo['id'], :full_result => true)
-          repo['errata'] = index(:errata, 'repository_id' => repo['id'], :full_result => true)
-        end
-
-        json = export_json(cvv, repositories)
+        json = export_json(export_json_options)
         create_tar(cvv, repositories, json)
         return HammerCLI::EX_OK
       end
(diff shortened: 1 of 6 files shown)


lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -287,18 +287,31 @@ class ExportCommand < HammerCLIForeman::Command
...
+        composite = cv["composite"]
+        export_json_options = { :cvv => cvv }
 
-        repositories.each do |repo|
-          repo['packages'] = index(:packages, 'repository_id' => repo['id'], :full_result => true)
-          repo['errata'] = index(:errata, 'repository_id' => repo['id'], :full_result => true)
+        if composite
+          components = cv['components']
+          export_json_options[:component_cvvs] = \
+            components.collect { |component| component['name'] }
```
ehelms: I'd rather see this as:

```
export_json_optionsp[:component_cvvs] = components.collect do | component|
  component['name']
end
```

if having to split over multiple lines due to line length.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -287,18 +287,31 @@ class ExportCommand < HammerCLIForeman::Command
...
+          export_json_options[:repositories] = []
+        else
+          export_json_options[:component_cvvs] = nil
```
ehelms: I think we could omit this if not a composite

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -412,19 +428,44 @@ def execute
...
-          export_json['minor'],
-          repos_units(export_json['repositories'])
-        )
+
+        if export_json['composite_components']
+          composite_version_ids = export_json['composite_components'].map {|component| find_local_component_id(component) }
+
+          resource_content_views.call(:update,
+                                      'id' => cv['id'], 'component_ids' => composite_version_ids)
+          publish(cv['id'], export_json['major'], export_json['minor'], nil)
```
ehelms: You could update the publish method signature to set `repo_units = nil` as a default and drop the trailing nil here.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -412,19 +428,44 @@ def execute
...
+        end
         return HammerCLI::EX_OK
       end
 
+      def find_local_component_id(component_from_export)
+        *name, version = component_from_export.split(' ')
+        name = name.join(' ')
+        existing_component_cv = content_view(name, options['option_organization_id'])
+        found_composite_version = \
+          existing_component_cv['versions'].select { |v| v['version'] == version }
```
ehelms: See similar comment above about preferring `do-end` style.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -412,19 +428,44 @@ def execute
...
+            export_json['major'],
+            export_json['minor'],
+            repos_units(export_json['repositories'])
+          )
+        end
         return HammerCLI::EX_OK
       end
 
+      def find_local_component_id(component_from_export)
+        *name, version = component_from_export.split(' ')
```
ehelms: What does this use of `*` and two return parameters result in? An example would be helpful to wrap my brain around this bit of code.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -412,19 +428,44 @@ def execute
...
+        name = name.join(' ')
+        existing_component_cv = content_view(name, options['option_organization_id'])
+        found_composite_version = \
+          existing_component_cv['versions'].select { |v| v['version'] == version }
+        if found_composite_version.empty?
+          raise _("unable to find CV version %{cvv} on system. Please ensure it " \
```
ehelms: `Unable` -- uppercase it

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -412,19 +428,44 @@ def execute
...
-          cv['id'],
-          export_json['major'],
-          export_json['minor'],
-          repos_units(export_json['repositories'])
-        )
+
+        if export_json['composite_components']
+          composite_version_ids = export_json['composite_components'].map {|component| find_local_component_id(component) }
+
+          resource_content_views.call(:update,
```
ehelms: You can just use `call` directly.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -412,19 +428,44 @@ def execute
...
+        if found_composite_version.empty?
+          raise _("unable to find CV version %{cvv} on system. Please ensure it " \
+                  "is already imported." % {'cvv' => component_from_export})
+        end
+        found_composite_version.first['id']
+      end
+
+      def resource_content_views
+        HammerCLIForeman.foreman_resource(:content_views)
+      end
```
ehelms: This is unneeded due to the ApipieHelper module

lib/hammer_cli_katello/content_view_version.rb
beav: Basically it says "put everything but the last thing in the array into `name`, and the last thing in `version`. For example:

```ruby
irb(main):001:0> *foo, bar = [1,2,3,4,5]
=> [1, 2, 3, 4, 5]
irb(main):002:0> bar
=> 5
irb(main):003:0> foo
=> [1, 2, 3, 4]
```
When we fetch the components from the CV, they are in "name + version number" format, and I want to split the name (which may have spaces) off from the version.

lib/hammer_cli_katello/content_view_version.rb
akofink: I think the time has come to implement `update` in the helper.

beav:
testing of this is blocked by https://projects.theforeman.org/issues/25318

beav:
@akofink I moved some of the "un-hammerish" import/export parsing code into a helper module, to try to improve readability

beav:
testing is unblocked by merged patch https://patch-diff.githubusercontent.com/raw/Katello/katello/pull/7786.patch, thanks @sjha4 !

beav:
@chris1984 if you dont mind, can you give the updated patch another test? If that goes OK, we can merge and create issues for usability improvements

cfouant:
ACK from me, great work @beav!

beav:
@ehelms @cfouant @chris1984 @akofink thanks all for the reviews!

PR #607, opened 2018-12-13
Fixes #25682 - Update export tar name to cvname-versionid

chris1984:
results of new output:

```
[root@centos7-katello-nightly ~]# ls
export-animals-2.tar  export-animals-3.tar  ssl-build
```

### DIFF
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -311,12 +311,12 @@ module HammerCLIKatello
         end
 
         json = export_json(export_json_options)
-        create_tar(cvv, repositories, json)
+        create_tar(cv, cvv, repositories, json)
         return HammerCLI::EX_OK
       end
 
-      def create_tar(cvv, repositories, json)
-        export_prefix = "export-#{cvv['id']}"
+      def create_tar(cv, cvv, repositories, json)
+        export_prefix = "export-#{cv['label']}-#{cvv['id']}"
         export_file = "#{export_prefix}.json"
         export_repos_tar = "#{export_prefix}-repos.tar"
         export_tar = "#{export_prefix}.tar"
--- a/test/functional/content_view/version/export_test.rb
+++ b/test/functional/content_view/version/export_test.rb
@@ -26,7 +26,8 @@ describe 'content-view version export' do
     ex = api_expects(:content_views, :show)
     ex.returns(
       'id' => '4321',
-      'composite' => false
+      'composite' => false,
+      'label' => 'cv'
     )
 
     ex = api_expects(:repositories, :show).with_params('id' => '2')
@@ -50,9 +51,9 @@ describe 'content-view version export' do
     File.stubs(:exist?).with('/var/log/hammer/hammer.log._copy_').returns(false)
 
     Dir.expects(:chdir).with("/var/lib/pulp/published/yum/https/repos/").returns(true)
-    Dir.expects(:mkdir).with('/tmp/exports/export-5').returns(0)
+    Dir.expects(:mkdir).with('/tmp/exports/export-cv-5').returns(0)
     Dir.expects(:chdir).with('/tmp/exports').returns(0)
-    Dir.expects(:chdir).with('/tmp/exports/export-5').returns(0)
+    Dir.expects(:chdir).with('/tmp/exports/export-cv-5').returns(0)
 
     result = run_cmd(@cmd + params)
     assert_equal(HammerCLI::EX_OK, result.exit_code)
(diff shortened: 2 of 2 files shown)


chris1984:
@beav I tried a few things to get the tests to pass but no go, when you get a chance can you take a look please?

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -316,7 +316,8 @@ def execute
       end
 
       def create_tar(cvv, repositories, json)
-        export_prefix = "export-#{cvv['id']}"
+        cv_name = cvv['name'].gsub(/\s.+/, '')
```
beav: is the cvv label available here? the label doesn't have spaces but I don't remember if it's passed in on the cvv hash

chris1984: changing the code to this:

`export_prefix = "export-#{cvv['label']}-#{cvv['id']}"`

Resulted in this:

`export--2.tar`

I tried with both 'label' and 'Label'

akofink: CVVs don't have names. They have versions and ids.

akofink: You're going to have to pass in the CV as well as the CVV to this method to achieve what you want.

chris1984: @akofink 

Something like this?

```
        cv = show(:content_views, 'id' => cvv['content_view_id'])
        export_prefix = "export-#{cv['name']}-#{cvv['id']}"
```

That got me the expected result.

beav: the 'name' field on the cvv var is from ` cvv = show(:content_view_versions, 'id' => options['option_id'])`. I'm guessing its just pulling in the CV name from that api call.

However, it's probably better to pass in both cvv and cv into `create_tar`, just so its more clear what's happening. `cv` might also have a label so you don't need to use the gsub :cat:

akofink: I think you might be able to access it with `cvv[:content_view][:label]` from looking at the [API response](https://github.com/Katello/katello/blob/master/app/views/katello/api/v2/content_view_versions/base.json.rabl#L28), but I could be wrong.

chris1984:
@akofink hmm its pulling the right name on my nightly and looks right, but the tests are still failing. Is this due to me not calling a correct param from the fake api in the tests?

akofink:
The error is because the chdir is ["Expected once, invoked twice"](https://travis-ci.org/Katello/hammer-cli-katello/jobs/468062586#L631). If you expect it only once, then there's a bug. Otherwise if you expect it twice, you can specify that in the tests with `Dir.expects(:mkdir).twice.with('/tmp/exports/export-cv-5').returns(0)`

chris1984:
@akofink everything is green and tests are updated :) I realized I had to grab the label param from the content view show api call.

beav:
I didn't test this but the code looks good. Just to confirm, this works correctly when the CV name has spaces?

chris1984:
> I didn't test this but the code looks good. Just to confirm, this works correctly when the CV name has spaces?

Let me test real quick

chris1984:
@beav it works:

```
hammer> content-view list
----------------|---------------------------|---------------------------|-----------|---------------------|---------------
CONTENT VIEW ID | NAME                      | LABEL                     | COMPOSITE | LAST PUBLISHED      | REPOSITORY IDS
----------------|---------------------------|---------------------------|-----------|---------------------|---------------
2               | animals                   | animals                   |           | 2018/12/13 15:30:40 | 1             
1               | Default Organization View | Default_Organization_View |           | 2018/12/13 15:17:16 |               
3               | i like spaces             | i_like_spaces             |           | 2018/12/14 18:00:31 | 1             
----------------|---------------------------|---------------------------|-----------|---------------------|---------------
```
```
hammer> content-view version list
---|-------------------------------|---------|-----------------------
ID | NAME                          | VERSION | LIFECYCLE ENVIRONMENTS
---|-------------------------------|---------|-----------------------
3  | animals 2.0                   | 2.0     | Library               
4  | i like spaces 1.0             | 1.0     | Library               
2  | animals 1.0                   | 1.0     |                       
1  | Default Organization View 1.0 | 1.0     | Library               
---|-------------------------------|---------|-----------------------
```
```
[root@centos7-katello-nightly ~]# ls
export-i_like_spaces-4.tar  ssl-build
[root@centos7-katello-nightly ~]# tar -xvf export-i_like_spaces-4.tar 
export-i_like_spaces-4/
export-i_like_spaces-4/export-i_like_spaces-4.json
export-i_like_spaces-4/export-i_like_spaces-4-repos.tar
```

PR #610, opened 2018-12-18
Fixes #25728 - Show Recurring Logic Id

sjha4:
Sync Plan list/info should also show the recurring logic id


akofink:
You know this only covers list, right?

akofink:
I mean, you only added it to `ListCommand`, not `InfoCommand`. You can open a new PR for `InfoCommand`.

PR #611, opened 2019-01-03
[WIP] Fixes #25775 - create content view if it doesn't exist

beav:
Previously, the content view had to be created ahead of time before
content view versions were imported.

This patch creates the content view if it doesn't already exist. Note
that it only creates the CV with name and repos, it does not do
activation keys. Composites are handled as well.

### DIFF
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -402,6 +402,15 @@ module HammerCLIKatello
 
         export_json = read_json(import_tar_params)
         cv = content_view(export_json['name'], options['option_organization_id'])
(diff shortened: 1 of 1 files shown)


akofink:
WIP for tests only, or are you adding more code?

beav:
> WIP for tests only, or are you adding more code?

tests and I want to move `find_library_repo` to import export helper.

akofink:
:hammer:

akofink:
Still working, or should we close this as stale?

chris1984:
@akofink ill close this one and cherrypick his pr so he gets credit for it, it is still an rfe that needs to be addressed. Beav said I could take it over

PR #612, opened 2019-01-08
Fixes #25805 - Display correct value for redhat repo

sjha4:
**Issue:**
Redhat repository field displays as false even for redhat enabled repos.

Depends on: https://github.com/Katello/katello/pull/7913


beav:
https://github.com/Katello/katello/pull/7913 is merged

PR #613, opened 2019-01-09
Fixes #25813 Docker tag info in man pages very vague

melcorr:
Previously, Docker tag info in man page was very ambiguous, leading to confusion. 

This makes it more clear.

### DIFF
--- a/lib/hammer_cli_katello/repository.rb
+++ b/lib/hammer_cli_katello/repository.rb
@@ -219,7 +219,8 @@ module HammerCLIKatello
       option "--publish-via-http", "ENABLE", _("Publish Via HTTP"),
              :attribute_name => :option_unprotected,
              :format => HammerCLI::Options::Normalizers::Bool.new
-      option "--docker-tag", "TAG", _("Container Image tag")
+      option "--docker-tag", "TAG",
+              _("Optional custom Container Image tag to specify the value of the Docker digest")
       option "--docker-digest", "DIGEST", _("Container Image manifest digest")
 
       def execute


lib/hammer_cli_katello/repository.rb
```diff
@@ -219,7 +219,7 @@ class UpdateCommand < HammerCLIKatello::UpdateCommand
       option "--publish-via-http", "ENABLE", _("Publish Via HTTP"),
              :attribute_name => :option_unprotected,
              :format => HammerCLI::Options::Normalizers::Bool.new
-      option "--docker-tag", "TAG", _("Container Image tag")
+      option "--docker-tag", "TAG", _("Optional custom Container Image tag to specify the value of the Docker digest")
```
akofink: Rubocop fails because the line is too long.

```suggestion
      option "--docker-tag", "TAG",
        _("Optional custom Container Image tag to specify the value of the Docker digest")
```

melcorr: Ah! Will fix.

lib/hammer_cli_katello/repository.rb
```diff
@@ -219,7 +219,7 @@ class UpdateCommand < HammerCLIKatello::UpdateCommand
       option "--publish-via-http", "ENABLE", _("Publish Via HTTP"),
              :attribute_name => :option_unprotected,
              :format => HammerCLI::Options::Normalizers::Bool.new
-      option "--docker-tag", "TAG", _("Container Image tag")
+      option "--docker-tag", "TAG", _("Optional Container Image tag to specify Docker digest")
```
akofink: Do you not want to match the documentation wording?

melcorr: I thought that the line was too long? Sorry this is the first time submitting a PR for this project. I think my original line is better, but if it fails validation, what can I do?

akofink: I showed in my first comment how to achieve what you want. You can just continue the statement on the next line, like this:

```suggestion
      option "--docker-tag", "TAG",
        _("Optional custom Container Image tag to specify the value of the Docker digest")
```

melcorr: I'm so thick!!

melcorr:
I squashed all my commits to get rid of my ill-formatted messages.

PR #616, opened 2019-01-14
Fixes #25857 - Make export dir always absolute

chris1984:
Before:

```
hammer content-view version export --export-dir test1 --id 2
tar: ExportedCV/export-2/export-2-repos.tar: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
```

With code change:
```
[root@dhcp129-218 test]# ll
total 0
drwxr-xr-x. 2 root root 6 Jan 14 11:17 test1
[root@dhcp129-218 test]# pwd
/tmp/test

[root@dhcp129-218 test]# ll -R
.:
total 0
drwxr-xr-x. 2 root root 26 Jan 14 11:18 test1

./test1:
total 152
-rw-r--r--. 1 root root 153600 Jan 14 11:18 export-5.tar
```

### DIFF
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -320,12 +320,13 @@ module HammerCLIKatello
         export_file = "#{export_prefix}.json"
         export_repos_tar = "#{export_prefix}-repos.tar"
         export_tar = "#{export_prefix}.tar"
+        export_dir = File.expand_path(options['option_export_dir'].to_s)
 
-        Dir.mkdir("#{options['option_export_dir']}/#{export_prefix}")
+        Dir.mkdir("#{export_dir}/#{export_prefix}")
 
         if repositories
           Dir.chdir(PUBLISHED_REPOS_DIR) do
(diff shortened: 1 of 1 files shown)


lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -320,12 +320,15 @@ def create_tar(cv, cvv, repositories, json)
         export_file = "#{export_prefix}.json"
         export_repos_tar = "#{export_prefix}-repos.tar"
         export_tar = "#{export_prefix}.tar"
+        # rubocop:disable Style/UnneededInterpolation
+        export_dir = File.expand_path("#{options['option_export_dir']}")
```
akofink: Why disable rubocop here instead of following the rule?

```suggestion
        export_dir = File.expand_path(options['option_export_dir'])
```

chris1984:
Thanks, @akofink sorry about that I was trying to get tests to work locally on my repo and forgot to change it before I committed it. Fixed, and now working on trying to make the tests green.

chris1984:
@akofink updated and good to go.

PR #624, opened 2019-01-24
Fixes #25919 - Add error message when import cv is nil

chris1984:
Before:

```ruby
Could not import the content view.:
  Error: undefined method `[]' for nil:NilClass
```

After:

```ruby
[ERROR 2019-01-24T20:45:33 Exception] Error: The Content View 'zoo' is not present on this server, please create the Content View and try the import again.
Could not import the content view:
  Error: The Content View 'zoo' is not present on this server, please create the Content View and try the import again.
```

Added the disable/enable on the abc cop because I was getting this:
```ruby
 lib/hammer_cli_katello/content_view_version.rb:396:7: C: Assignment Branch Condition size for execute is too high. [40.31/38]
```

Removed the period on line `381` because it made the error message look odd:

`before: Could not import the content view.:`

`after: Could not import the content view:`

### DIFF
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -378,7 +378,7 @@ module HammerCLIKatello
       command_name "import"
 
       success_message _("Content view imported.")
-      failure_message _("Could not import the content view.")
+      failure_message _("Could not import the content view")
 
       option "--organization-id", "ORGANIZATION_ID", _("Organization numeric identifier")
       option(
@@ -393,6 +393,7 @@ module HammerCLIKatello
 
       build_options
 
+      # rubocop:disable Metrics/AbcSize
       def execute
         unless File.exist?(options['option_export_tar'])
           raise _("Export tar #{options['option_export_tar']} does not exist.")
@@ -403,6 +404,11 @@ module HammerCLIKatello
 
         export_json = read_json(import_tar_params)
         cv = content_view(export_json['name'], options['option_organization_id'])
+
+        if cv.nil?
+          raise _("The Content View '#{export_json['name']}' is not present on this server,"\
+          " please create the Content View and try the import again.")
+        end
 
         if export_json['composite_components']
           composite_version_ids = export_json['composite_components'].map do |component|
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -378,7 +378,7 @@ class ImportCommand < HammerCLIForeman::Command
       command_name "import"
 
       success_message _("Content view imported.")
-      failure_message _("Could not import the content view.")
+      failure_message _("Could not import the content view")
```
ehelms: Why drop the period?

chris1984: Removed the period on line 381 because it made the error message look odd:

`before: Could not import the content view.:`

`after: Could not import the content view:`

Plus it looks like the other failures do not have them but the success do:

https://github.com/Katello/hammer-cli-katello/blob/master/lib/hammer_cli_katello/content_view_version.rb#L74

https://github.com/Katello/hammer-cli-katello/blob/master/lib/hammer_cli_katello/content_view_version.rb#L87

https://github.com/Katello/hammer-cli-katello/blob/master/lib/hammer_cli_katello/content_view_version.rb#L135

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -422,6 +428,7 @@ def execute
         end
         return HammerCLI::EX_OK
       end
+      # rubocop:enable Metrics/AbcSize
```
johnpmitsch: If there is an easy way to avoid disabling the cops by breaking up the method, that would be better IMO, but if it makes things more convoluted and less readable, I think its fine to disable the cop. (I'm not sure how strict this repo is about things like this)

akofink: This method is pretty scary looking IMO :)

chris1984: After talking with @akofink I made a redmine issue to refactor the execute method after the 6.5 beta stuff is in. Here is the the redmine to track it.

https://projects.theforeman.org/issues/25931

chris1984:
@ehelms can I get merge I need this for my other prs to pass. because of the cop

lib/hammer_cli_katello/content_view_version.rb
chris1984: Added to Kanboard as well.

PR #625, opened 2019-01-28
Fixes #25932 - Prevent reimport of same CV version

chris1984:
This PR checks to see if the CV version being imported has already been imported or the latest version is higher than the version being imported, and if not gives the user an error that is more friendly and exists:

Before:

The import would not check and it would try to kick off a sync of the repos on the import side and then kick off a publish of the content view which would be bad and could cause issues with stuck tasks etc.

After:
```
[vagrant@centos7-devel hammer_cli_katello]$ hammer content-view version import --organization-id 2 --export-tar /tmp/export-7.tar
Could not import the content view.:
  Error: The latest version (7.1) of the Content View pizza is greater than the version you are trying to import (7.1)
```

### DIFF
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -404,23 +404,19 @@ module HammerCLIKatello
 
         export_json = read_json(import_tar_params)
         cv = content_view(export_json['name'], options['option_organization_id'])
-
-        if cv.nil?
-          raise _("The Content View '#{export_json['name']}' is not present on this server,"\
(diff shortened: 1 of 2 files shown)


chris1984:
Error checking still works after refactoring the checks in the execute method to their own.

```
Could not import the content view:
  Error: The Content View zoo is not present on this server, please create the Content View and try the import again.

Could not import the content view:
  Error: The latest version (5.1) of the Content View 'zoo' is greater or equal to the version you are trying to import (5.1)
```

chris1984:
@ehelms can I have a review, please?

PR #629, opened 2019-02-12
Refs #23204 - Consistent puppet environment naming in hammer

ofedoren:
Moved the `puppet-environment` name resolution to `hammer-cli-foreman`, see: https://github.com/theforeman/hammer-cli-foreman/pull/401


ofedoren:
@akofink, right, thanks! Updated.

akofink:
Please ping me when https://github.com/theforeman/hammer-cli-foreman/pull/401 is merged. Thanks!

ofedoren:
@akofink, there are some extra changes. Same logic as in https://github.com/theforeman/hammer-cli-foreman/pull/401: changes `--environment` options to `--lifecycle-environment`.

akofink:
Please ping me when the hammer-cli-foreman PR merges so we can merge this.

akofink:
Also, you may want to make a community post about this change. It may break some people's hammer scripts, [including our bats tests](https://github.com/theforeman/forklift/search?q=--environment&unscoped_q=--environment).

ofedoren:
@akofink, I've just noticed a slight inconsistency: why is here (https://github.com/Katello/hammer-cli-katello/blob/d059ffb207a258322e60f7b4fad95a8bbd6c51bf/lib/hammer_cli_katello/content_view_version.rb#L147) expanded `environments` option? API documentation for that endpoint (DELETE /katello/api/content_view_versions/:id ) says that `id` is only parameter for it. The same for `activation-key` command's endpoints (though `environment` options were there I've removed them being based on apidocs in a separate commit).

Please, can you confirm if it is a right approach or I'm missing something?

akofink:
@ofedoren You may change the option expansion code, but know that content view version IDs are a bit tricky to resolve - you need the content view ID AND either the version (i.e. 3.5) OR the lifecycle environment (i.e. Development). I don't believe there is a test currently for delete with ID resolution (or at all, for that matter :/ )

ofedoren:
@akofink, the [hammer-cli-foreman PR](https://github.com/theforeman/hammer-cli-foreman/pull/401) is merged.

chris1984:
rekicked the tests

ekohl:
Looks like tests started to fail after this: https://ci.theforeman.org/job/hammer-cli-katello-master-source-release/10/

akofink:
Yes, you cannot successfully run the hammer-cli-katello master branch against the old released version of hammer-cli-foreman. If you run everything from master, it works fine.

PR #630, opened 2019-02-20
Fixes #26091 - Require product-id on cv update

parthaa:
This commit fixes the following call
hammer content-view update
--id=6 --organization-id=4  --repositories=goo

It fixes it by requiring the user to specify the product-id when the
repositories flag is used.

### DIFF
--- a/test/functional/content_view/update_test.rb
+++ b/test/functional/content_view/update_test.rb
@@ -57,7 +57,8 @@ module HammerCLIKatello
         cmd = %W(content-view update --id=#{id}\
                  --repository-ids=#{repo_id} --product=#{product_name})
         result = run_cmd(cmd)
-        assert(result.err[/--organization-id, --organization, --organization-label is required/],
+        expected_error = "--organization-id, --organization, --organization-label is required"
+        assert(result.err.include?(expected_error),
                "Organization option requirements are validated")
       end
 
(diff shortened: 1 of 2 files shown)


parthaa:
> I think we should add another option validator to improve the error message. Also, please add tests for this scenario.

Addressed

```
hammer content-view update --id 2 --repositories repo1 --product test
Could not update the content view:
  Error: At least one of options --organization-id, --organization, --organization-label is required.
  
  See: 'hammer content-view update --help'.
```

test/functional/content_view/update_test.rb
```diff
@@ -48,6 +48,30 @@ module HammerCLIKatello
...
+      it 'requires product if repository names are provided' do
+        id = 122
+        repo_id = 100
+        organization_id = 5
+        api_expects_no_call
+        cmd = %W(content-view update --id=#{id}\
+                 --repositories=#{repo_id} --organization-id=#{organization_id})
+        result = run_cmd(cmd)
+        assert(result.err[/--product-id, --product is required/],
+               "Product option requirements are validated")
```
akofink: This string is shown in the test output when this test fails. IMO it should describe what went wrong, [like this](https://github.com/Katello/hammer-cli-katello/blob/49b48b3c1d431da10c254e3c59e62552cb909122/test/functional/package/list_test.rb#L34).

test/functional/content_view/update_test.rb
```diff
@@ -48,6 +48,30 @@ module HammerCLIKatello
...
+      it 'requires organization if product name is supplied along with repository' do
+        id = 122
+        repo_id = 100
+        product_name = "foo"
+        api_expects_no_call
+        cmd = %W(content-view update --id=#{id}\
+                 --repository-ids=#{repo_id} --product=#{product_name})
+        result = run_cmd(cmd)
+        assert(result.err[/--organization-id, --organization, --organization-label is required/],
+               "Organization option requirements are validated")
```
akofink: Same here.

PR #639, opened 2019-04-08
[CP] Picks for 0.16.1 (Katello 3.10.1)

jturel:


jturel:
@akofink any thoughts on why there are so many test failures? Maybe I'm missing something that needs to be picked over here?

akofink:
The failures are coming from hammer-cli 0.16 (it should be 0.15.z). @mbacovsky hammer-cli-foreman 0.15.z tests also fail with hammer-cli 0.16, but [the hammer-cli-foreman gemspec shows `'hammer_cli', '>= 0.15.1'`](https://github.com/theforeman/hammer-cli-foreman/blob/0.15-stable/hammer_cli_foreman.gemspec#L28). Do you know why the hammer-cli version isn't limited to `'< 0.16'` as well?

Adding the following to hammer-cli-katello's gemspec fixes the issue:

```rb
gem.add_dependency 'hammer_cli', '>= 0.15.1', '< 0.16'
```

akofink:
Ok, I've figured it out. We need to do two things to make this PR green:

1. Get [12fc9d6](https://github.com/theforeman/hammer-cli-foreman/commit/12fc9d63c1150b0303956c91d2b40531f7f93271) in a hammer-cli-foreman 0.15.z release @mbacovsky? Currently hammer-cli-foreman 0.15.z is broken when testing with hammer-cli 0.16.
1. Cherry-pick [bf661b5](https://github.com/Katello/hammer-cli-katello/commit/bf661b5c718a280531525dd3f563c0b88462dc42) in hammer-cli-katello

jturel:
@akofink I updated https://github.com/theforeman/hammer-cli-foreman/pull/419 which in combination with the two reverted commits on this PR will pass tests locally.

I think it's OK to bump these things off of the release because we're getting into the territory of having to update hammer_cli_foreman and hammer_cli to support whatever had changed validations-wise. Your fix addressed enabling Red Hat repos via hammer, but the UI may still be used so I wouldn't consider it critical. What do you think? If you agree I can tidy up this PR and we should have a green build once the above is merged.

jturel:
[test]

PR #640, opened 2019-04-08
Fixes #26257 - Use pages on package/errata index to prevent t/o

chris1984:
Export not having timeout and showing correct:
```
hammer -d content-view version export --id 2 --export-dir /tmp

[DEBUG 2019-04-08T15:33:30 API] Params: {
    "repository_id" => 5,
              :page => 193,
          :per_page => 50
}
[DEBUG 2019-04-08T15:33:30 API] Headers: {
    :params => {
        "repository_id" => 5,
                  :page => 193,
              :per_page => 50
    }
}

[root@exportpatch tmp]# du -sh /tmp/export-timeout-2.tar 
45G	/tmp/export-timeout-2.tar
```
Import machine, importing correctly:
```
hammer -d content-view version import --organization-id 1 --export-tar /var/lib/pulp/katello-export/export-timeout-2.tar

                                 :vary => "Accept-Encoding",
                     :content_encoding => "gzip",
                       :content_length => "534",
                         :content_type => "application/json; charset=utf-8"
}
[.....................................................................................................................................................................................................................................................] [100%]
```
Screenshot of repos looking good:

RHEL - https://nimb.ws/YLe8pw

RHSCL - https://nimb.ws/xHC4yg

Screenshot of CV published and imported:

https://nimb.ws/P4sYQm

Client able to install `screen` package from timeout cv:
```
Installed:
  screen.x86_64 0:4.1.0-0.25.20120314git3c2946.el7                                                                                                                                                                                                            

Complete!

Repo-id      : rhel-7-server-rpms/7Server/x86_64
Repo-name    : Red Hat Enterprise Linux 7 Server (RPMs)
Repo-revision: 1554752590
Repo-updated : Mon Apr  8 19:43:10 2019
Repo-pkgs    : 23,918
Repo-size    : 34 G
Repo-baseurl : https://importpatch.toledo.satellite.lab.eng.rdu2.redhat.com/pulp/repos/Default_Organization/Library/timeout/content/dist/rhel/server/7/7Server/x86_64/os
Repo-expire  : 1 second(s) (last: Mon Apr  8 19:53:20 2019)
  Filter     : read-only:present
Repo-filename: /etc/yum.repos.d/redhat.repo
```

### DIFF
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -291,6 +291,7 @@ module HammerCLIKatello
         cv = show(:content_views, 'id' => cvv['content_view_id'])
 
         composite = cv["composite"]
+
         export_json_options = { :cvv => cvv }
 
         if composite
@@ -304,11 +305,8 @@ module HammerCLIKatello
           puppet_check(cvv)
           check_repo_type(repositories)
           check_repo_download_policy(repositories)
+          collect_packages(repositories)
 
-          repositories.each do |repo|
-            repo['packages'] = index(:packages, 'repository_id' => repo['id'], :full_result => true)
-            repo['errata'] = index(:errata, 'repository_id' => repo['id'], :full_result => true)
-          end
           export_json_options[:repositories] = repositories
         end
 
--- a/test/functional/content_view/version/export_test.rb
+++ b/test/functional/content_view/version/export_test.rb
@@ -22,7 +22,6 @@ describe 'content-view version export' do
       'content_view' => {'name' => 'cv'},
       'content_view_id' => 4321,
       'puppet_modules' => []
-
     )
 
     ex = api_expects(:content_views, :show)
@@ -39,7 +38,11 @@ describe 'content-view version export' do
       'content_type' => 'yum',
       'backend_identifier' => 'Default_Organization-Library-Test_Repo',
       'relative_path' => 'Default_Organization/Library/Test_Repo',
-      'library_instance_id' => '1'
+      'library_instance_id' => '1',
+      'content_counts' => {
+        'rpm' => 1,
+        'erratum' => 1
+      }
     )
 
     api_expects(:repositories, :show).with_params('id' => '1').returns(
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -8,7 +8,7 @@
 
 # Offense count: 7
 Metrics/AbcSize:
-  Max: 38
+  Max: 39
 
 # Offense count: 1
 Metrics/PerceivedComplexity:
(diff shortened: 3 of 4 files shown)


lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -262,6 +262,7 @@ class LegacyExportCommand < HammerCLIKatello::SingleResourceCommand
       end
     end
 
+    # rubocop:disable Metrics/ClassLength
```
akofink: This is a sign. If we just disable Rubocop every time we want to do something nasty, why use it at all? I feel the same about the other `rubocop:disable` places. We should break up this class.

chris1984:
@jyejare here is the pr if you want to test it

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -286,11 +287,13 @@ class ExportCommand < HammerCLIForeman::Command
 
       build_options
 
+      # rubocop:disable Metrics/MethodLength
```
akofink: We should absolutely break up this method; It's way too long and complicated.

lib/hammer_cli_katello/content_view_version.rb
chris1984: I will do that on the next bug since I have to add some more code to the class, this time I found out I didnt need to disable that since I was still working on the pr. Will mark this one as resolved.

lib/hammer_cli_katello/content_view_version.rb
chris1984: Yeah I agree, I will do that on the next push with the test fixes.

chris1984:
@akofink updated the loop on the repo hash to collect packages/errata to it is own method and moved it to the cv_helper. Let me know if you want me to keep it in the main file and I can move it back, rubocop disables are gone now. 

Also if you have any insight into why the test main export test is failing I would appreciate it, I fixed the other 2 but this one is failing still. I am guessing it is due to the for loop I have going?

test/functional/content_view/version/export_test.rb
```diff
@@ -22,7 +22,6 @@
       'content_view' => {'name' => 'cv'},
       'content_view_id' => 4321,
       'puppet_modules' => []
-
```
akofink: This should fix the test:

```diff
diff --git a/test/functional/content_view/version/export_test.rb b/test/functional/content_view/version/export_test.rb
index 8ea0929..afe317b 100644
--- a/test/functional/content_view/version/export_test.rb
+++ b/test/functional/content_view/version/export_test.rb
@@ -38,7 +38,11 @@ describe 'content-view version export' do
       'content_type' => 'yum',
       'backend_identifier' => 'Default_Organization-Library-Test_Repo',
       'relative_path' => 'Default_Organization/Library/Test_Repo',
-      'library_instance_id' => '1'
+      'library_instance_id' => '1',
+      'content_counts' => {
+        'rpm' => 1,
+        'erratum' => 1
+      }
     )

     api_expects(:repositories, :show).with_params('id' => '1').returns(
```

PR #643, opened 2019-05-03
Bump to 0.19-master for use with Katello nightly/3.13

akofink:


lib/hammer_cli_katello/lifecycle_environment_name_resolvable.rb
```diff
@@ -17,7 +17,7 @@ def get_options(_defined_options, result)
     def option_sources
       sources = super
       sources.find_by_name('IdResolution').insert_relative(
-        :after,
+        :before,
```
akofink: @mbacovsky should this cause any problem? I had to do this to get the lifecycle environment resolver to happen before the puppet environment resolver.

PR #644, opened 2019-05-21
Refs #25878 - Rename tags

ofedoren:
Requires https://github.com/theforeman/hammer-cli/pull/300.

### DIFF
--- a/lib/hammer_cli_katello/output/formatters.rb
+++ b/lib/hammer_cli_katello/output/formatters.rb
@@ -2,8 +2,8 @@ module HammerCLIKatello
   module Output
     module Formatters
       class ChecksumFormatter < HammerCLI::Output::Formatters::FieldFormatter
-        def tags
-          [:screen]
+        def required_features
+          [:rich_text]
         end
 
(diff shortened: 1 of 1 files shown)


akofink:
I'm confused why this change is necessary. If we continue to supply the `tags` method, won't it get mapped to features in [`tag_to_feature`](https://github.com/theforeman/hammer-cli/pull/300/files#diff-baa81df89d6eb513c95777be662dc4d7R13) by [`required_features`](https://github.com/theforeman/hammer-cli/pull/300/files#diff-25287b89c6737c88215955d7f09cedcbR36)? And [`:screen` maps to `:rich_text`](https://github.com/theforeman/hammer-cli/pull/300/files#diff-baa81df89d6eb513c95777be662dc4d7R9).

ofedoren:
Yes it will, but also there will be deprecation message which says that we want to get rid of this method. Not sure for now if it will be printed in logs (then this change is not necessary) or it will be shown to all the users.

mbacovsky:
Related PRs were merged. Merging.

PR #649, opened 2019-06-04
Fixes #26946 - Fix check_repo_download_policy

patilsuraj767:
Not sure why repo['library_instance_id'] was used, repo['library_instance_id'] returns the nil vaule.

### DIFF
--- a/lib/hammer_cli_katello/cv_import_export_helper.rb
+++ b/lib/hammer_cli_katello/cv_import_export_helper.rb
@@ -45,10 +45,10 @@ module HammerCLIKatello
 
     def check_repo_download_policy(repositories)
       non_immediate = repositories.select do |repo|
-        show(:repositories, 'id' => repo['library_instance_id'])['download_policy'] != 'immediate'
+        show(:repositories, 'id' => repo['id']) if repo['download_policy'] != 'immediate'
       end
       unless non_immediate.empty?
-        non_immediate_names = repositories.collect { |repo| repo['name'] }
+        non_immediate_names = non_immediate.collect { |repo| repo['name'] }
         msg = <<~MSG
           All exported repositories must be set to an immediate download policy and re-synced.
           The following repositories need action:


lib/hammer_cli_katello/cv_import_export_helper.rb
```diff
@@ -45,10 +45,10 @@ def check_repo_type(repositories)
 
     def check_repo_download_policy(repositories)
       non_immediate = repositories.select do |repo|
-        show(:repositories, 'id' => repo['library_instance_id'])['download_policy'] != 'immediate'
+        show(:repositories, 'id' => repo['id']) if repo['download_policy'] != 'immediate'
```
akofink: Hmm... which version of Katello are you using? `library_instance_id` was added [here](https://github.com/Katello/katello/commit/21af1ea72870b4ecedbd81cd3688d391a949d418). It's not yet released. We should keep this on master, but we need to fix released versions of hammer that use this `library_instance_id`, which looks to be included starting in hammer-cli-katello 0.15.0 https://github.com/Katello/hammer-cli-katello/commit/949663022c7ed41a9ff43d7f59b9f8432838cbf5.

patilsuraj767: @akofink  I have observed this in Satellite 6.5 
Katello version - katello-3.10.0-0.6.rc1.el7sat.noarch
hammer - tfm-rubygem-hammer_cli_katello-0.16.0.11-1.el7sat.noarch

I have noticed one more thing, that this works in case of any other content view other than  "Default Organization View". It returns the value for library_instance_id for other content-views.

akofink: Speaking with the authors of this export feature, it seems exporting the default organization view is unsupported and should be disallowed to prevent user confusion in the future. I think it's safe to close this PR, but thanks for submitting!

akofink:
Exporting the default organization view is unsupported and should be disallowed to prevent user confusion in the future. I'm closing this PR in favor of a future PR to disallow exporting the default organization view.

PR #651, opened 2019-06-12
Fixes #27039 - Content View version Export Condition Fix

patilsuraj767:
"repositories" return the empty array if there is no repo added in the content view. So we should check the length of the array instead checking if it is set.

### DIFF
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -314,7 +314,10 @@ module HammerCLIKatello
         if repositories&.any? || cv['composite']
           create_tar(cv, cvv, repositories, json)
         else
-          puts "Ensure the content view version '#{cvv['name']}' has at least one repository."
+          msg = <<~MSG
+            Ensure the content view version '#{cvv['name']}' has at least one repository.
+          MSG
+          raise _(msg)
         end
         return HammerCLI::EX_OK
       end
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -324,7 +324,7 @@ module HammerCLIKatello
 
         Dir.mkdir("#{export_dir}/#{export_prefix}")
 
-        if repositories.any?
+        if repositories&.any?
           Dir.chdir(PUBLISHED_REPOS_DIR) do
             repo_tar = "#{export_dir}/#{export_prefix}/#{export_repos_tar}"
             repo_dirs = []
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -322,7 +322,6 @@ module HammerCLIKatello
         export_tar = "#{export_prefix}.tar"
         export_dir = File.expand_path(options['option_export_dir'].to_s)
 
-
         if repositories&.any?
           Dir.mkdir("#{export_dir}/#{export_prefix}")
           Dir.chdir(PUBLISHED_REPOS_DIR) do
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -324,7 +324,7 @@ module HammerCLIKatello
 
         Dir.mkdir("#{export_dir}/#{export_prefix}")
 
-        if !repositories.empty?
+        if repositories.any?
           Dir.chdir(PUBLISHED_REPOS_DIR) do
             repo_tar = "#{export_dir}/#{export_prefix}/#{export_repos_tar}"
             repo_dirs = []
@@ -336,7 +336,7 @@ module HammerCLIKatello
             `tar cvfh #{repo_tar} #{repo_dirs.join(" ")}`
           end
         else
-          puts "Ensure the content view '#{cvv['name']}' has atleast one repository."
+          puts "Ensure the content view '#{cvv['name']}' has at least one repository."
         end
 
         Dir.chdir("#{export_dir}/#{export_prefix}") do
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -324,7 +324,7 @@ module HammerCLIKatello
 
         Dir.mkdir("#{export_dir}/#{export_prefix}")
 
-        if repositories.length > 0
+        if !repositories.empty?
           Dir.chdir(PUBLISHED_REPOS_DIR) do
             repo_tar = "#{export_dir}/#{export_prefix}/#{export_repos_tar}"
             repo_dirs = []
@@ -336,7 +336,7 @@ module HammerCLIKatello
             `tar cvfh #{repo_tar} #{repo_dirs.join(" ")}`
           end
         else
-          puts "Please verify that content view version id - #{cvv['id']} contents atleast one repository"
+          puts "Ensure the content view '#{cvv['name']}' has atleast one repository."
         end
 
         Dir.chdir("#{export_dir}/#{export_prefix}") do
--- a/lib/hammer_cli_katello/content_view_version.rb
+++ b/lib/hammer_cli_katello/content_view_version.rb
@@ -324,7 +324,7 @@ module HammerCLIKatello
 
         Dir.mkdir("#{export_dir}/#{export_prefix}")
 
-        if repositories
+        if repositories.length > 0
           Dir.chdir(PUBLISHED_REPOS_DIR) do
             repo_tar = "#{export_dir}/#{export_prefix}/#{export_repos_tar}"
             repo_dirs = []
@@ -335,6 +335,8 @@ module HammerCLIKatello
 
             `tar cvfh #{repo_tar} #{repo_dirs.join(" ")}`
           end
+        else
+          puts "Please verify that content view version id - #{cvv['id']} contents atleast one repository"
         end
 
         Dir.chdir("#{export_dir}/#{export_prefix}") do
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -335,6 +335,8 @@ def create_tar(cv, cvv, repositories, json)
 
             `tar cvfh #{repo_tar} #{repo_dirs.join(" ")}`
           end
+        else
+          puts "Please verify that content view version id - #{cvv['id']} contents atleast one repository"
```
akofink: s/atleast/at least/

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -324,7 +324,7 @@ def create_tar(cv, cvv, repositories, json)
 
         Dir.mkdir("#{export_dir}/#{export_prefix}")
 
-        if repositories
+        if repositories.length > 0
```
akofink: `if repositories.any?`

lib/hammer_cli_katello/content_view_version.rb
chris1984: s/contents/contains

patilsuraj767:
@chris1984 @akofink Sorry I have checked the you update after pushing the commit. 

I have modified the whole message as it was going beyond 100 char and test was failing. 

Also changed   if repositories.length > 0  To   if !repositories.empty? let me if that is fine.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -324,7 +324,7 @@ def create_tar(cv, cvv, repositories, json)
 
         Dir.mkdir("#{export_dir}/#{export_prefix}")
 
-        if repositories
+        if !repositories.empty?
```
akofink: This can be `if repositories.any?`

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -335,6 +335,8 @@ def create_tar(cv, cvv, repositories, json)
 
             `tar cvfh #{repo_tar} #{repo_dirs.join(" ")}`
           end
+        else
+          puts "Ensure the content view '#{cvv['name']}' has atleast one repository."
```
akofink: There's still a typo here: s/atleast/at least/

lib/hammer_cli_katello/content_view_version.rb
patilsuraj767: Ok I will change it. Thanks.

lib/hammer_cli_katello/content_view_version.rb
patilsuraj767: Yes, will change this too.

chris1984:
@patilsuraj767 I will start testing this tomorrow and work on letting you know why the tests are failing. Assigning to me

patilsuraj767:
@chris1984 Sure, Thanks.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -324,7 +324,7 @@ def create_tar(cv, cvv, repositories, json)
 
         Dir.mkdir("#{export_dir}/#{export_prefix}")
 
-        if repositories
+        if repositories.any?
```
akofink: `s/repositories.any?/repositories&.any?/` will fix your tests

patilsuraj767: @akofink Thanks that worked

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -335,6 +335,8 @@ def create_tar(cv, cvv, repositories, json)
 
             `tar cvfh #{repo_tar} #{repo_dirs.join(" ")}`
           end
+        else
+          puts "Ensure the content view '#{cvv['name']}' has at least one repository."
```
akofink: Hmm... I just realized another issue. The test that was failing was for composite content view export (in which repositories is nil). This message doesn't make sense in that context.

patilsuraj767:
I think we will need to put the below two "Dir" blocks inside the if condition, because it always create the .tar file, even if the command is failing with the error.
Let me know your thoughts on this @chris1984 @akofink  

```
       Dir.chdir("#{export_dir}/#{export_prefix}") do
          File.open(export_file, 'w') do |file|
            file.write(JSON.pretty_generate(json))
          end
        end

        Dir.chdir(export_dir) do
          `tar cf #{export_tar} #{export_prefix}`
          FileUtils.rm_rf(export_prefix)
        end

```

chris1984:
> I think we will need to put the below two "Dir" blocks inside the if condition, because it always create the .tar file, even if the command is failing with the error.
> Let me know your thoughts on this @chris1984 @akofink
> 
> ```
>        Dir.chdir("#{export_dir}/#{export_prefix}") do
>           File.open(export_file, 'w') do |file|
>             file.write(JSON.pretty_generate(json))
>           end
>         end
> 
>         Dir.chdir(export_dir) do
>           `tar cf #{export_tar} #{export_prefix}`
>           FileUtils.rm_rf(export_prefix)
>         end
> ```

If we do that the tests will prob break and need updating but yeah imho creating a tar with nothing in on error is a bad experience for the user

lib/hammer_cli_katello/content_view_version.rb
akofink: IMO you can fix the composite issue by checking `elsif !cv['composite']`

chris1984:
Thanks @patilsuraj767 ill test again, @akofink does this look good? If so ill merge after testing

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -335,17 +334,22 @@ def create_tar(cv, cvv, repositories, json)
...
-        Dir.chdir(export_dir) do
-          `tar cf #{export_tar} #{export_prefix}`
-          FileUtils.rm_rf(export_prefix)
+          Dir.chdir(export_dir) do
+            `tar cf #{export_tar} #{export_prefix}`
+            FileUtils.rm_rf(export_prefix)
+          end
+
+        elsif cv['composite']
+          puts "Cannot export composite content view."
```
akofink: Definitely not true - composite CVs are exported without repositories, just the json.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -322,9 +322,8 @@ def create_tar(cv, cvv, repositories, json)
         export_tar = "#{export_prefix}.tar"
         export_dir = File.expand_path(options['option_export_dir'].to_s)
 
-        Dir.mkdir("#{export_dir}/#{export_prefix}")
-
-        if repositories
```
akofink: You need to check if the content view is not composite here, and completely skip this block if it is.

patilsuraj767:
@akofink My bad. I thought we cannot export the content view as there is no repository in it.
I have changed the code again and added the test cases back. As far as I have tested, now we are able to export the composite content view and also, there is no tar file created if the content view version export fails. Please review it again.

test/functional/content_view/version/export_test.rb
```diff
@@ -90,7 +90,7 @@
 
     File.expects(:exist?).with('/usr/share/foreman').returns(true)
     File.stubs(:exist?).with('/var/log/hammer/hammer.log._copy_').returns(false)
-
+
```
akofink: Please remove the vestigial white space changes.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -311,7 +311,11 @@ def execute
         end
 
         json = export_json(export_json_options)
-        create_tar(cv, cvv, repositories, json)
+        if repositories&.any? or cv['composite']
```
akofink: Use `||` instead of `or`, due to [precedent issues](http://ruby-doc.org/core-2.6.3/doc/syntax/precedence_rdoc.html). I'm surprised rubocop didn't complain about this.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -325,6 +329,7 @@ def create_tar(cv, cvv, repositories, json)
         Dir.mkdir("#{export_dir}/#{export_prefix}")
 
         if repositories
```
akofink: I think you still need to check if `repositories&.any?` here, as an empty repositories array (`[]`) will still be exported.

lib/hammer_cli_katello/content_view_version.rb
patilsuraj767: rubocop has complained about it. I just thought to correct it later, to avoid unnecessary commit.

lib/hammer_cli_katello/content_view_version.rb
patilsuraj767: I think at this point empty repositories array will not be present because we are checking for empty repositories in the previous function. 

```
if repositories&.any? or cv['composite']
     create_tar(cv, cvv, repositories, json)
else
     puts "Ensure the content view version '#{cvv['name']}' has at least one repository."
end
```

akofink: Could repositories still be `[]` though if it's a composite CV?

patilsuraj767: @akofink I do not think that repository can be `[]` in case of composite CV. 
Repository return `[]` only if any content view is published without adding any repository in it. 
In can of composite CV, I have observed that repository variable itself is not set. But anyways its good change it to `repositories&.any?`. I will make the changes and send the commit again.

patilsuraj767:
@akofink Added the test case.

chris1984:
Will retest and if it works ill merge. The code looks good :+1:

PR #652, opened 2019-06-14
Refs #26135 - Allow hammer to send off content-type

chris1984:
Tests are failing because the Katello PR is not in yet, once that is in I will come back to this. Adding in now for a review from @akofink and then once the Katello PR is in I will get the tests fixed up etc. Setting to  Waiting on Contrib.


lib/hammer_cli_katello/repository.rb
```diff
@@ -436,12 +435,14 @@ def update_content_upload(upload_id, repo_id, file)
...
       def import_uploads(uploads, opts = {})
+        rpm_type = option_content_type
         publish_repository = opts.fetch(:last_file, false)
         sync_capsule = opts.fetch(:last_file, false)
         params = {:id => get_identifier,
                   :uploads => uploads,
                   publish_repository: publish_repository,
-                  sync_capsule: sync_capsule
+                  sync_capsule: sync_capsule,
+                  content_type: rpm_type,
```
akofink: Not sure why we need the `rpm_type` variable instead of just passing in `option_content_type`.

chris1984:
@akofink  So looking at the test failures would you be opposed to an if option_content_type? here for this and else just pass in the params as normal?

https://github.com/Katello/hammer-cli-katello/blob/27f157975646be274fe83d68242e42d9df344d05/lib/hammer_cli_katello/repository.rb#L440

Tests are failing it looks like because we are expecting a param to be passed and since it is optional i think that is the best route but wanted to ask you first

akofink:
I don't think that will fix your tests. You need to include the updated apipie cache.

akofink:
I think this needs a rebase

vijay8451:
@chris1984 I tried to fetch this one but not working due to conflict .

chris1984:
> @chris1984 I tried to fetch this one but not working due to conflict .

Updated

akofink:
You should probably only update the apipie cache for Katello 3.13. These tests aren't run against 3.12. https://github.com/Katello/hammer-cli-katello/blob/master/test/test_helper.rb#L20

chris1984:
@akofink yep I am going to, also working on adding a test for the content type too,

PR #656, opened 2019-06-29
Fixes #27187 - Fixing Content View version Export for default organiz…

vision2910:
Fixed the content-view version export for default organization view by putting a check to ensure no nil library instance for the repository is passed through while checking the repo download policy.

### DIFF
--- a/lib/hammer_cli_katello/cv_import_export_helper.rb
+++ b/lib/hammer_cli_katello/cv_import_export_helper.rb
@@ -45,7 +45,9 @@ module HammerCLIKatello
 
     def check_repo_download_policy(repositories)
       non_immediate = repositories.select do |repo|
-        show(:repositories, 'id' => repo['library_instance_id'])['download_policy'] != 'immediate'
+        unless repo['library_instance_id'].nil?
+          show(:repositories, 'id' => repo['library_instance_id'])['download_policy'] != 'immediate'
+        end
       end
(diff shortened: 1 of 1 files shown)


vision2910:
stumbled upon https://projects.theforeman.org/issues/26952 and learn that exporting the default organization view is unsupported and should be disallowed to prevent user confusion in the future. In that case, the provided PR becomes invalid.

chris1984:
@vision2910 it is on the backlog, we need to get it done at some point, so right now we just want to error out until we get the code added. The issue is open for taking :) Otherwise I will get to when I get some time.

vision2910:
raised a PR for https://projects.theforeman.org/issues/26952.

@chris1984 @akofink sounds like a good idea for testing this PR's functionality. I will add a test for this case.

mccun934:
I tested this, works fine, exported the default org view without issues.

chris1984:
Thanks @vision2910

vision2910:
Thanks @mccun934, @chris1984, @akofink

akofink:
Merged without tests? :( :see_no_evil:

chris1984:
I will make another pr with updated tests me-- :(

vision2910:
> I will make another pr with updated tests me-- :(

added the test https://github.com/Katello/hammer-cli-katello/pull/659

PR #660, opened 2019-07-08
Refs #27213 - Add update filter test

chris1984:


chris1984:
@akofink I was wrong looks like all the filter tests are in the content view folder, so just adding a test for the other pr I made on updating a filter description.

PR #662, opened 2019-07-10
Fixes #27261 - export-legacy should not be described as deprecated

wbclark:
`export-legacy` method has different functionality from `export` and should not be described as deprecated


chris1984:
@akofink yeah there was talk about removing it but I am not sure where that went. Since I have not seen anything new around that topic I am fine with this change too since we are still supporting both ways. Marking as approved. Thanks, @wbclark :beers:

wbclark:
@akofink @chris1984 Thank you for the quick turnaround on this one! Export-legacy does provide different functionality than the new export method, for example the new export method will fail when trying to export any CV version that contains non-RPM content, additionally the format of the exported content is different and some users have processes that are specifically dependent on that formatting. Therefore it's important to continue to support export-legacy.

PR #663, opened 2019-07-16
fixes #24170 - SysPurpose on content host info

sseelam2:


lib/hammer_cli_katello/host_extensions.rb
```diff
@@ -90,6 +89,15 @@ module HostExtensions
           field :id, _('Id')
           field :name, _('Name')
         end
+
+        label _('System Purpose') do
```
akofink: Perhaps we should nest this under `Subscription Information`?

sseelam2: Yes. That makes sense. Thanks. :+1:

PR #664, opened 2019-07-16
fixes #27230 - SysPurpose act key info

sseelam2:


test/functional/activation_key/info_test.rb
```diff
@@ -0,0 +1,29 @@
+require_relative '../test_helper'
+require 'hammer_cli_katello/associating_commands'
+
+describe 'activation-key info' do
+  before do
+    @cmd = %w(activation-key info)
+  end
+
+  it "includes katello attributes" do
+    params = ['--id=1', '--organization-id=1']
```
akofink: No need for `--organization-id` if you are using `--id`

test/functional/activation_key/update_test.rb
```diff
@@ -19,7 +19,7 @@ module HammerCLIKatello
           p['id'] == 6 && p['organization_id'] == 1
         end
 
-        run_cmd(%w(activation-key update --organization org1 --name ak --auto-attach false))
+        run_cmd(%w(activation-key update --organization org1 --name ak --auto-attach false ))
```
akofink: Was this intentional?

sseelam2: Oh no. Sorry. That's a mistake.

test/functional/activation_key/info_test.rb
sseelam2: ah alright. I'll make that change. Thanks.

test/functional/activation_key/info_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    params = ['--id=1']
+    ex = api_expects(:activation_keys, :show, "Activation key") do |par|
+      par['id'] == 1
+    end
+
+    json_file = File.join(File.dirname(__FILE__), 'data', 'activation_key.json')
+    ex.returns(JSON.parse(File.read(json_file)))
+
+    result = run_cmd(@cmd + params)
+    # rubocop:disable Style/WordArray
```
akofink: This is not required and is causing the build to fail.

sseelam2: alright. I have removed it.

PR #666, opened 2019-07-22
Fixes #27374 - Add export-default to export default cv like CDN

chris1984:


chris1984:
@mccun934 if you want to test it?

@akofink do you know of a good way to stub out /var/lib/pulp/published/yum/https/repos/ for the test? it is failing since rsync cant cd to the directory.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -267,6 +267,35 @@ def request_params_hosts(params)
...
+      validate_options do
+        option(:option_export_dir).required
+      end
+
+      build_options
+
+      def execute
+        export_dir = options['option_export_dir']
+
+        `rsync -avL #{PUBLISHED_REPOS_DIR} #{export_dir}`
```
akofink: You could use `system()` here and stub the `system` method in your test.

chris1984:
@akofink thanks for the pointer :) How does this look?

Once @mccun934 comments on the true testing then I think we can merge

test/functional/content_view/version/default_export_test.rb
```diff
@@ -0,0 +1,24 @@
...
+  it "performs export" do
+    params = [
+      '--export-dir=/tmp/default'
+    ]
+
+    File.expects(:exist?).with('/usr/share/foreman').returns(true)
+    File.stubs(:exist?).with('/var/log/hammer/hammer.log._copy_').returns(false)
+
+    Dir.expects(:mkdir).with('/tmp/default').returns(0)
+    Kernel.expects(:system).with("rsync -avL /var/lib/pulp/published/yum/https/repos/ /tmp/default")
```
akofink: Yep! Looks right to me. :)

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -267,6 +267,36 @@ def request_params_hosts(params)
...
+      end
+
+      build_options
+
+      def execute
+        export_dir = options['option_export_dir']
+
+        Dir.mkdir(export_dir)
+        Kernel.system("rsync -avL #{PUBLISHED_REPOS_DIR} #{export_dir}")
+        return HammerCLI::EX_OK
```
akofink: You could check the rsync return value. From the system docs: "system returns true if the command gives zero exit status, false for non zero exit status." :man_shrugging:

chris1984:
@akofink updated based on the last comment, also added another test to make it fail and check for that:

Here are the results of the test from my devel box:
```bash
Good:

[vagrant@centos7-katello-devel ~]$ hammer content-view version export-default --export-dir /home/vagrant/test
Default content view export is available in /home/vagrant/test

[vagrant@centos7-katello-devel ~]$ ll /home/vagrant/test/
total 4
drwxr-x---. 4 vagrant apache 57 Jul 23 18:43 Default_Organization
-rw-r--r--. 1 vagrant apache 20 Jul 23 18:43 listing

Fail:

[vagrant@centos7-katello-devel ~]$ hammer content-view version export-default --export-dir /root
rsync: change_dir#1 "/root" failed: Permission denied (13)
rsync error: errors selecting input/output files, dirs (code 3) at main.c(614) [Receiver=3.1.2]
Could not export the content view at /root
```
Mike said he won't be able to test till next week, so up to you if you want to merge/ack or let it sit. I am fine with either since it is not a urgent issue.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -267,6 +267,42 @@ def request_params_hosts(params)
...
+      end
+
+      build_options
+
+      def execute
+        export_dir = options['option_export_dir']
+
+        Dir.mkdir(export_dir) unless Dir.exist?(export_dir)
+        result = Kernel.system("rsync -aL #{PUBLISHED_REPOS_DIR} #{export_dir}")
+        if result == true
```
akofink: This is not normal in Ruby, but I can understand why you must explicitly check for `true`. From the ruby docs:

> system returns true if the command gives zero exit status, false for non zero exit status. Returns nil if command execution fails

I think it will return `nil` only if the `rsync` command isn't found on the system, which is relevant.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -267,6 +267,42 @@ def request_params_hosts(params)
...
+          puts "#{success_message} #{export_dir}"
+          return HammerCLI::EX_OK
+        else
+          puts "#{failure_message} #{export_dir}"
```
akofink: You should be able to use `output.print_error` and `output.print_message`, defined [here](https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/abstract.rb#L163) and [here](https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/output/output.rb#L8).

Be sure to wrap your strings in `_("my string here")` for translations

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -267,6 +267,42 @@ def request_params_hosts(params)
...
+          return HammerCLI::EX_CANTCREAT
```
akofink: The `return` keywords here and line 298 are not required (optional to remove)

chris1984:
@akofink updated. I removed the puts and replaced with the output method, removed the failure/success messages since they are not used and also removed the return statements.

Here are the updated tests:
```bash

Good:

[vagrant@centos7-katello-devel hammer_cli_katello-0.18.0]$ hammer content-view version export-default --export-dir /home/vagrant/test
Default content view export is available at /home/vagrant/test
[vagrant@centos7-katello-devel hammer_cli_katello-0.18.0]$ ll /home/vagrant/test
total 4
drwxr-x---. 4 vagrant apache 57 Jul 23 18:43 Default_Organization
-rw-r--r--. 1 vagrant apache 20 Jul 23 18:43 listing

Bad:
[vagrant@centos7-katello-devel hammer_cli_katello-0.18.0]$ hammer content-view version export-default --export-dir /root
rsync: change_dir#1 "/root" failed: Permission denied (13)
rsync error: errors selecting input/output files, dirs (code 3) at main.c(614) [Receiver=3.1.2]
Could not export the default content view at /root
```

lib/hammer_cli_katello/content_view_version.rb
chris1984: I agree I was having some issues with tests and it working correctly until I did it this way. Not the best way but sadly it will have to do for now until we push export into its own module and look at a better way to export etc.

PR #679, opened 2019-08-07
Bump master to 0.20

chris1984:


akofink:
We're getting some build failures... It'll need some investigation. I will try to take a look as well.

chris1984:
> Is there a new apipie cache we can put in test/data and bump
> 
> https://github.com/Katello/hammer-cli-katello/blob/a971d702932e784e8e0eb77b20d1432119836c0e/test/test_helper.rb#L20
> 
> ?

@akofink i opened this one up https://github.com/Katello/hammer-cli-katello/pull/680 I will see why it is failing

PR #698, opened 2019-09-23
Fixes #27778 - Org command no longer shows deprication warning

ofedoren:
@akofink, could you or someone from your team please ensure that this will not break commands when `environments` should be mapped to `lifecycle-environments`? I think some commands that use `--lifecycle-environments` as an option will be enough.


ofedoren:
@akofink, thanks for directing! Now it should work as expected. But if you prefer to do more testing, please do :)

PR #596, opened 2018-10-12
Refs #3036 - Ping command

ofedoren:
requires https://github.com/theforeman/hammer-cli-foreman/pull/394 and https://github.com/Katello/katello/pull/7751 to be merged first

PR #699, opened 2019-09-26
Fixes #27947 - Allow duplicate file uploads

sjha4:
This change allows uploading duplicate files to pulp3.

When creating upload_request we pass the file checksum. In katello, we check if there is a content_unit with the same checksum. If yes, we return the content_unit_id to hammer. If hammer receives a content_unit_id in the result, we skip uploading the file and proceed to importing the same content_unit to the repository.

The associated changes need the katello PR - https://github.com/Katello/katello/pull/8364.


chris1984:
We should check this with Pulp2 as well since it is changing the stuff I did for the SRPM feature work.

sjha4:
@akofink , @chris1984 : So I need to probably add new apidoc to the data folder. I think I need to follow the steps here https://github.com/theforeman/hammer-cli-foreman/blob/master/test/data/README.md but I am confused if this has to happen all in the foreman directory? Also what would be the folder the apidocs go into?

sjha4:
> We should check this with Pulp2 as well since it is changing the stuff I did for the SRPM feature work.

Yes. Locally, I tested this with rpm without the --content-type flag and srpm with --content-type=srpm.

To explain the change around content_type here, now the content type is also passed while creating the upload. The reason being that in pulp3, we need to call type specific plugin to figure if there's a duplicate content unit already uploaded. If there is a duplicate, instead of going through the upload, we directly import the existing content unit into the repository.

akofink:
@sjha4 It's steps 3 through 7 in those instructions. Instead of hammer-cli-foreman test/data folder, you would use hammer-cli-katello's test/data folder and update the 3.14 apipie cache at https://github.com/Katello/hammer-cli-katello/tree/master/test/data/3.14.

chris1984:
@akofink do we need this merged so the tests will pass? https://github.com/theforeman/hammer-cli-foreman/pull/394

ofedoren:
@chris1984, if I'm not mistaken, there is only one test (the second one which fails) that requires https://github.com/Katello/hammer-cli-katello/pull/596 (thus https://github.com/theforeman/hammer-cli-foreman/pull/394)  to pass. The first test that fails is not related to that PR.

chris1984:
@sjha4 emailed you a patch with what I did to get it working while testing on my devel box looking at the api data that is coming back.

akofink:
Wait on https://github.com/Katello/katello/pull/8364 before merge

sjha4:
Merged the related katello PR. Thanks @ianballou , @chris1984 , @akofink !

chris1984:
ACK, Katello pr is merged, merging. Test failure is unrelated and passes on master so overriding failed test pr merge policy based on that

PR #704, opened 2019-10-29
Bump to 0.20.0 for Katello 3.14

chris1984:


chris1984:
@akofink updated, thanks for catching that.

PR #711, opened 2019-12-10
Fixes #28178 - Correct http_proxy search by name

chris1984:
This PR addresses an issue where when you search http-proxy by name it returns all of the entries instead of just 1 which will cause a failure.

This is because of https://github.com/Katello/hammer-cli-katello/blob/165de784748030762a52f6642e9951001bd50c6e/lib/hammer_cli_katello/id_resolver.rb#L52, so the search_options (or search parameters) are wrongly built.

@jjeffers want to test it?

### DIFF
--- a/lib/hammer_cli_katello/foreman_search_options_creators.rb
+++ b/lib/hammer_cli_katello/foreman_search_options_creators.rb
@@ -87,5 +87,9 @@ module HammerCLIKatello
     def create_users_search_options(options, mode = nil)
       create_search_options_without_katello_api(options, api.resource(:users), mode)
     end
+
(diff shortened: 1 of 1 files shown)


jjeffers:
@chris1984 Tested this locally with an updated hammer-foreman client and this PR. I was able to add a repository with an assigned HTTP proxy. The assigned proxy had a name that also matched a second HTTP proxy, but correctly used the one that matched first instead of throwing an error.

chris1984:
Merging based on @akofink ack and @jjeffers testing. Thanks All and @ofedoren for the pointer in hammer-foreman.

PR #714, opened 2020-01-07
Fixes #28293 - rename gpg command

jjeffers:
This PR marks the `gpg` command as deprecated, and duplicates the implementation to respond to `content-credentials`.

New tests were added as a skeleton for more comprehensive testing for the `content-credentials` API resource. No tests existed for the `gpg` command prior.

To test, for example:
```
hammer content-credentials list --organization-id <your org id here>
```
Any `gpg` command should not be usable, but return a deprecation warning via STDERR. The same subcommands exist for `content-credentials`.

### DIFF
--- a/lib/hammer_cli_katello.rb
+++ b/lib/hammer_cli_katello.rb
@@ -43,9 +43,20 @@ module HammerCLIKatello
                                           'hammer_cli_katello/organization'
                                          )
 
-  HammerCLI::MainCommand.lazy_subcommand("gpg", _("Manipulate GPG Key actions on the server"),
+  HammerCLI::MainCommand.lazy_subcommand("gpg",
+                                         _("Manipulate GPG Key actions on the server"),
                                          'HammerCLIKatello::GpgKeyCommand',
-                                         'hammer_cli_katello/gpg_key'
+                                         'hammer_cli_katello/gpg_key',
+                                         :warning =>
+  _('The gpg sub-command is deprecated and will be removed in one of the future versions.' \
+    ' Please use the content-credentials command instead.')
+                                        )
+
+  HammerCLI::MainCommand.lazy_subcommand("content-credentials",
+                                         _("Manipulate content credentials (i.e. GPG Keys)' \
+                                          'on the server"),
+                                         'HammerCLIKatello::ContentCredentialCommand',
+                                         'hammer_cli_katello/content_credential'
                                         )
 
   HammerCLI::MainCommand.lazy_subcommand("lifecycle-environment",
--- a/test/functional/content_credentials/info_test.rb
+++ b/test/functional/content_credentials/info_test.rb
@@ -1,0 +1,50 @@
+require File.join(File.dirname(__FILE__), '../test_helper')
+require File.join(File.dirname(__FILE__), '../product/product_helpers')
+
+def expect_credentials_search(org_id, name, id)
+  api_expects(:content_credentials, :index, 'List content credentials')
+    .with_params('name' => name, 'organization_id' => org_id)
+    .returns(index_response([{'id' => id}]))
+end
+
+describe "get content-credentials info" do
+  include ProductHelpers
+
+  before do
+    @cmd = %w(content-credentials info)
+  end
+
+  let(:org_id) { 1 }
+  let(:product_id) { 2 }
+  let(:key_id) { 3 }
+
+  it "Shows information about a content credental" do
+    params = ["--organization-id=#{org_id}", "--name=test_key"]
+
+    expect_credentials_search(org_id, 'test_key', key_id)
+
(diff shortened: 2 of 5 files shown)


lib/hammer_cli_katello.rb
```diff
@@ -43,9 +43,19 @@ def self.exception_handler_class
                                           'hammer_cli_katello/organization'
                                          )
 
-  HammerCLI::MainCommand.lazy_subcommand("gpg", _("Manipulate GPG Key actions on the server"),
+  HammerCLI::MainCommand.lazy_subcommand("gpg",
+                                         _("Manipulate GPG Key actions on the server"),
                                          'HammerCLIKatello::GpgKeyCommand',
-                                         'hammer_cli_katello/gpg_key'
+                                         'hammer_cli_katello/gpg_key',
+                                         :warning =>
```
akofink: There's actually a `deprecated` key you can pass here, like this: https://github.com/Katello/hammer-cli-katello/blob/d70bd7ad6d02ba2b9e9d0d67b0d5e51109a4b760/lib/hammer_cli_katello/command_extensions/lifecycle_environments.rb#L8

lib/hammer_cli_katello.rb
```diff
@@ -43,9 +43,19 @@ def self.exception_handler_class
...
+  _('gpg command is deprecated and will be removed in one of the future versions.' \
```
akofink:
```suggestion
  _('The gpg sub-command is deprecated and will be removed in one of the future versions.' \
```

lib/hammer_cli_katello.rb
```diff
@@ -43,9 +43,19 @@ def self.exception_handler_class
...
+    ' Please use content-credentials command instead.')
```
akofink:
```suggestion
    ' Please use the content-credentials command instead.')
```

lib/hammer_cli_katello.rb
```diff
@@ -43,9 +43,19 @@ def self.exception_handler_class
...
+                                        )
+
+  HammerCLI::MainCommand.lazy_subcommand("content-credentials",
+                                         _("Manipulate GPG Key actions on the server"),
```
akofink: GPG keys? Or will this API support non-gpg credentials now or in the future?

test/functional/content_credentials/list_test.rb
```diff
@@ -0,0 +1,68 @@
+require File.join(File.dirname(__FILE__), '../test_helper')
+require File.join(File.dirname(__FILE__), '../lifecycle_environment/lifecycle_environment_helpers')
+
+require 'hammer_cli_katello/content_view_puppet_module'
+
+describe 'listing repositories' do
```
akofink:
```suggestion
describe 'listing content credentials' do
```

test/functional/content_credentials/list_test.rb
```diff
@@ -0,0 +1,68 @@
...
+        "order" => nil
+      },
+      "results" => []
+    }
+  end
+
+  it "lists an organizations keys" do
+    params = ["--organization-id=#{org_id}"]
+
+    ex = api_expects(:content_credentials, :index, 'Organizations repositories list') do |par|
```
akofink: This will be very confusing unless we update the expectation description

```suggestion
    ex = api_expects(:content_credentials, :index, 'Content credentials list') do |par|
```

Check for other instances of 'repository' in this test file and change to content credentials accordingly.

lib/hammer_cli_katello.rb
jjeffers: @akofink I don't know the answer to that question. The original issue was just asking for a command rename. Who would be a good person to ask about the intended direction of that endpoint?

akofink: I'd say we can go ahead and change it to say something like this (exact wording up to you):

```suggestion
                                         _("Manipulate content credentials (i.e. GPG Keys) on the server"),
```

chris1984:
Looks like Andrew hit on the stuff I would have left comments on. Once @akofink comments are addressed I will test. Marked myself as assigned to remind myself to test once Andrew acks it.

test/functional/content_credentials/info_test.rb
```diff
@@ -0,0 +1,50 @@
...
+  let(:org_id) { 1 }
+  let(:product_id) { 2 }
+  let(:key_id) { 3 }
+
+  it "Shows information about a content credental" do
+    params = ["--organization-id=#{org_id}", "--name=test_key"]
+
+    expect_credentials_search(org_id, 'test_key', key_id)
+
+    ex = api_expects(:content_credentials, :show, "Get info") do |par|
```
akofink: We tend to prefer the newer `with_params` syntax for better error messages.

PR #716, opened 2020-01-22
Fixes #28833 - Add title resolver to organization resource

shiramax:


ofedoren:
@shiramax, LGTM and fixes the issue.

But rubocop isn't happy :(

shiramax:
@ofedoren done :)

test/functional/package/list_test.rb
```diff
@@ -86,8 +86,9 @@ module HammerCLIKatello
       it 'may be specified environment name no org fails' do
         api_expects_no_call
         r = run_cmd(%w(package list --environment Library))
```
akofink: Could we please add at least one test that uses this new option?

shiramax: @akofink 
I tried to find similar tests for id resolution and couldn't find it. 
can you give an example for a test that I can check?

akofink: [Here's one for srpm list](https://github.com/Katello/hammer-cli-katello/blob/3159a231e99d37d0703871f2d68a293f95b38626/test/functional/srpm/list_test.rb#L57) in which we're testing that organization_id is resolved from `--organization-label`. I'm not sure why we don't have any organization tests in hammer-cli-katello :/ You could create test/functional/organization/show_test.rb which requires organization_id but should be able to be resolved from --organization (name), --organization-label, and --organization-title.

akofink: I just looked at the redmine - why not add a test for that command so we can catch regressions in the future? `hammer organization delete --title <title name>`. That'd be the same as the show_test.rb I recommended above but in test/functional/organization/delete_test.rb - it requires the organization ID and must be resolved from name, label, or title.

shiramax: Done :) @akofink let me know if that's ok :)

akofink: Looks great! There is a rubocop failure for formatting.

We also have this test organization helpers at https://github.com/Katello/hammer-cli-katello/blob/master/test/functional/organization/organization_helpers.rb which we could use for these new tests, but I'm not going to block this PR on that.

shiramax: I don't mind using  the expect_organization_search just that the search I need to do is  {:search => "name = \"\"my_org\"\"", :per_page => 1000, :page => 1} and in expect_organization_search I can only send {"search": "name = \"my_org\""}

PR #720, opened 2020-01-27
Fixes #28294 - Create repo with ssl cert name

johnpmitsch:
This adds the ability to create a repository with a content credential names, not just ids

For example:

hammer repository create --product-id 3 --name my_repo --url https://example.com/myawesomerepo --content-type yum --ssl-client-key-name "my product key" --ssl-client-cert-name "my product cert" --ssl-ca-cert-name "my ca"

### DIFF
--- a/lib/hammer_cli_katello/repository.rb
+++ b/lib/hammer_cli_katello/repository.rb
@@ -201,7 +201,35 @@ module HammerCLIKatello
              :attribute_name => :option_unprotected,
              :format => HammerCLI::Options::Normalizers::Bool.new
 
+      option "--ssl-client-cert", "SSL_CLIENT_CERT",
+             _("Name of content credential containing client certificate"),
+             :attribute_name => :option_ssl_client_cert
+
+      option "--ssl-client-key", "SSL_CLIENT_KEY",
+             _("Name of content credential containing client key"),
+             :attribute_name => :option_ssl_client_key
+
+      option "--ssl-ca-cert", "SSL_CA_CERT",
+             _("Name of content credential containing CA certificate"),
+             :attribute_name => :option_ssl_ca_cert
+
       build_options :without => [:unprotected]
+
+      def request_params
+        params = super
+
+        params["ssl_client_cert_id"] ||= get_gpg_key_id("ssl_client_cert") if option_ssl_client_cert
+        params["ssl_client_key_id"] ||= get_gpg_key_id("ssl_client_key") if option_ssl_client_key
+        params["ssl_ca_cert_id"] ||= get_gpg_key_id("ssl_ca_cert") if option_ssl_ca_cert
+
+        params
+      end
+
+      def get_gpg_key_id(ssl_type)
+        options_clone = options.clone
+        options_clone["option_gpg_key_name"] = self.send("option_#{ssl_type}")
+        resolver.gpg_key_id(resolver.scoped_options("gpg_key", options_clone))
+      end
     end
 
     class UpdateCommand < HammerCLIKatello::UpdateCommand
--- a/test/functional/repository/create_test.rb
+++ b/test/functional/repository/create_test.rb
@@ -1,0 +1,51 @@
+require_relative '../test_helper'
+
+describe "create repository" do
+  let(:org_id) { 1 }
+  let(:product_id) { 2 }
+  let(:name) { "repo1" }
+  let(:content_type) { "yum" }
+
+  it 'with basic options' do
+    api_expects(:repositories, :create)
+      .with_params(
+        name: name,
+        product_id: product_id,
+        content_type: content_type)
+
+    command = %W(repository create --organization-id #{org_id} --product-id #{product_id}
+                 --content-type #{content_type} --name #{name})
+
+    assert_equal(0, run_cmd(command).exit_code)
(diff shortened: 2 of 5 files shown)


johnpmitsch:
To test:

1. Create 3 content credentials, these can be actual credentials, but for the purpose of testing out parameters (and not actually syncing w/ credentials) you can use a random string to fake out the content credential
2. checkout PR and run `FOREMAN_APIPIE_LANGS=en bundle exec rake apipie:cache` on the katello box and then run `hammer -r` on a hammer devel box or wherever you have hammer set up.
3. Create a repository with the content credential names passed in, for example: `hammer repository create --product-id 3 --name my_repo --url https://example.com/myawesomerepo --content-type yum --ssl-client-key-name "my product key" --ssl-client-cert-name "my product cert" --ssl-ca-cert-name "my ca"`

johnpmitsch:
I'm trying to write tests for this but I can't get the parameters to match the mocked API calls for some reason. 

I'm mostly going off examples in the codebase, if anyone has any ideas what is going wrong, I could use some guidance! :)

johnpmitsch:
> Looks like we added in a file called !

Thanks, just removed it!

chris1984:
Helping with the tests

test/functional/repository/create_test.rb
```diff
@@ -0,0 +1,41 @@
...
+        'content_type' => content_type,
+        'ssl_ca_cert_id' => 1,
+        'ssl_client_cert_id' => 2,
+        'ssl_client_key_id' => 3,
+        'name' => name)
+    end
+
+    command = %W(repository create --organization-id #{org_id} --product-id #{product_id}
+                 --content-type #{content_type} --name #{name} --ssl-client-cert-name test_cert
+                 --ssl-client-key-name test_key --ssl-ca-cert-name test_ca)
```
akofink: This doesn't quite match other options from other commands. The `-name` portion of an option is usually implied if we leave off the field, so it maybe should accept `--ssl-client-cert`, `--ssl-client-key`, `--ssl-ca-cert`.

test/functional/repository/create_test.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    command = %W(repository create --organization-id #{org_id} --product-id #{product_id}
+                 --content-type #{content_type} --name #{name})
+
+    assert_equal(0, run_cmd(command).exit_code)
+  end
+
+  it 'with ssl options by name' do
+    api_expects(:repository, :create, 'create a repository') do
+      with_params(
```
akofink: Instead of passing a block here, it's much easier to debug failed tests if we call `with_params` directly:

```rb
api_expects(:repository, :create, 'create a repository').with_params(...)
```

test/functional/repository/create_test.rb
```diff
@@ -0,0 +1,41 @@
...
+        'organization_id' => org_id,
+        'product_id' => product_id,
+        'content_type' => content_type,
+        'ssl_ca_cert_id' => 1,
+        'ssl_client_cert_id' => 2,
+        'ssl_client_key_id' => 3,
```
akofink: I suspect the test is failing because one of these ids is a string (i.e. `'1'`) rather than an int. It's difficult to tell though unless we change to using `with_params` as described above.

johnpmitsch:
Updated the tests and removed `-name` from the parameter

lib/hammer_cli_katello/repository.rb
```diff
@@ -201,7 +201,35 @@ class CreateCommand < HammerCLIKatello::CreateCommand
...
+             :attribute_name => :option_ssl_ca_cert
+
       build_options :without => [:unprotected]
+
+      def request_params
+        params = super
+
+        params["ssl_client_cert_id"] = get_gpg_key_id("ssl_client_cert") if option_ssl_client_cert
+        params["ssl_client_key_id"] = get_gpg_key_id("ssl_client_key") if option_ssl_client_key
+        params["ssl_ca_cert_id"] = get_gpg_key_id("ssl_ca_cert") if option_ssl_ca_cert
```
akofink: Maybe these should be memoized in case a user passes both ID and name? Usually we'd have an option_validation to ensure users cannot pass both ID and name.

chris1984:
Thanks @johnpmitsch

lib/hammer_cli_katello/repository.rb
johnpmitsch: That is a good point, I added memoization which means the ids will have precedent over the names, I think that is fine for this (hopefully) rare case of passing both names and IDs.

One thing that could help this area is mention the `ssl_*_id` params are content credentials in the API docs. the name parameters and id parameters don't match:
```
 --ssl-ca-cert SSL_CA_CERT                                         Name of content credential containing CA certificate
 --ssl-ca-cert-id SSL_CA_CERT_ID                                   Idenifier of the SSL CA Cert
```
 Its not super clear right now (or giving any more clarity to the parameter name) that the SSL CA Cert id is an id of a content credential. There is also a typo :)

I will open a Katello PR for this!

johnpmitsch: opened here: https://github.com/Katello/katello/pull/8546

PR #722, opened 2020-02-03
Fixes #28908 - Add pulp3 dep warning to repo create

chris1984:
With PR:

```bash
hammer repository create --name food --content-type puppet --product-id 1
Puppet and OSTree will no longer be supported in Katello 3.16
Repository created.

hammer repository create --name food1 --content-type ostree --product-id 1 --url 'http://foo.example.com'
Puppet and OSTree will no longer be supported in Katello 3.16
Repository created.
```

### DIFF
--- a/lib/hammer_cli_katello/repository.rb
+++ b/lib/hammer_cli_katello/repository.rb
@@ -222,7 +222,16 @@ module HammerCLIKatello
         params["ssl_client_key_id"] ||= get_gpg_key_id("ssl_client_key") if option_ssl_client_key
         params["ssl_ca_cert_id"] ||= get_gpg_key_id("ssl_ca_cert") if option_ssl_ca_cert
 
+        dep_warning
+
         params
+      end
+
+      def dep_warning
+        content = options.clone
+        if content['option_content_type'] == 'puppet' || content['option_content_type'] == 'ostree'
+          print_message _('Puppet and OSTree will no longer be supported in Katello 3.16')
+        end
       end
 
       def get_gpg_key_id(ssl_type)
--- a/test/functional/repository/create_test.rb
+++ b/test/functional/repository/create_test.rb
@@ -48,4 +48,36 @@ describe "create repository" do
 
     assert_equal(0, run_cmd(command).exit_code)
   end
+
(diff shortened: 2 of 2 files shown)


lib/hammer_cli_katello/repository.rb
```diff
@@ -222,9 +222,18 @@ def request_params
...
+        dep_warning
+
         params
       end
 
+      def dep_warning
+        content = options.clone
+        if content['option_content_type'] == 'puppet' || content['option_content_type'] == 'ostree'
+          print_message _('Puppet and OSTree will no longer be supported in Katello 3.16')
+        end
```
akofink: Should we add a test?

chris1984: You know it :)

chris1984:
@akofink updated, let me know if that looks good

jjeffers:
Probably not adding anything new other than to report that after applying this PR, I was able to see the deprecation warnings as @chris1984 mentioned above.

test/functional/repository/create_test.rb
```diff
@@ -48,4 +48,36 @@ def stub_gpg_key(gpg_key_name)
...
+      .with_params(
+        name: name,
+        product_id: product_id,
+        content_type: "puppet")
+
+    command = %W(repository create --organization-id #{org_id} --product-id #{product_id}
+                 --content-type puppet --name #{name})
+
+    assert('Puppet and OSTree will no longer be supported in Katello 3.16',
+           'Puppet and OSTree will no longer be supported in Katello 3.16')
```
akofink: This assertion seems incorrect. Same with the one on line 78. It should be `assert(<something that is true>, <error message if not true>)`, like `assert(food == 'salad', 'Food isn't salad!')`

jjeffers: Or `assert_equal <value1>, <value2>, message`

chris1984:
@akofink updated and it is testing correctly now, since i had to spend some time getting the formatting correctly. I added the Rubocop line disable because it is 4 chars over and if i try to move the text around the formatting of the assert just gets wonky, normally I would not do that but in this case it seems appropriate. The main code adheres to Rubocop.

PR #723, opened 2020-02-06
Fixes #28949 - added warning key in hash

patilsuraj767:
warning key was missing in the hash therefore it was returning null.


chris1984:
I'll take a look at the test failure

chris1984:
This passes locally on my machine, kicking off tests again.

PR #724, opened 2020-02-14
Bump to 0.21.0 for Katello 3.15

jturel:
Seeing if this builds before branching.


akofink:
You'll also probably need to include the updated Katello 3.15 apipie cache.

jturel:
> The base should be 0.21.z, not master. master should get bumped to `0.22-master`. I've just created the 0.21.z branch from master for you.

Yeah, I just wanted to see if the tests would run

jturel:
Thanks for pushing the branch!

jturel:
[test]

jturel:
@akofink any sense of these failures? I don't mind fixing them but need a better idea. Do they look like real failures or could something else be broken?

akofink:
I'm not exactly sure, but it seems like something wrong with the versions of hammer_cli and/or hammer_cli_foreman. I'd try running with master versions of all 3, and if it fails, try the latest release of all 3, then use git bisect to figure out when it broke. I think @chris1984 was going to figure out the broken tests on master here.

jturel:
This will be green once there's a new release of hammer-cli-foreman (probably 2.0.1)

ekohl:
Just linking https://github.com/theforeman/hammer-cli-foreman/pull/504 for completeness

jturel:
@akofink @chris1984 green!

PR #726, opened 2020-02-17
Bump version to 0.22-master

jturel:
Also added .vendor/ to gitignore since I install the bundle there


jturel:
@akofink  oops. this is the PR i meant to ask you about the test failures on, not the other one. do these looks like regressions or tests the need updating, or something else?

akofink:
With only 3 failures, yes it seems like regressions or tests that need updating, but I'm not sure which. Does it pass against the 3.15 api? You don't need to update the 3.15 apipie cache in this PR.

chris1984:
@jturel  Did you want me to try and fix the tests this week when I get time?

test/functional/activation_key/add_host_collection_test.rb
```diff
@@ -20,10 +20,12 @@ module HostCollection
         end
 
         it 'allows resolving host collection name' do
-          ex = api_expects(:host_collections, :index) do |p|
-            p['name'] == 'hc3' && p['organization_id'] == 1
+          2.times do
```
jturel: I'm concerned that this is a bug. I don't know why this API called be happening twice now

test/functional/content_view/add_content_view_version_test.rb
```diff
@@ -20,11 +20,13 @@ module HammerCLIKatello
     end
 
     it 'resolves content view version ID' do
-      ex = api_expects(:content_view_versions, :index) do |p|
-        p['content_view_id'] == 3 && p['version'] == '2.1'
+      2.times do
```
jturel: Ditto about this duplicated API call

jturel:
> With only 3 failures, yes it seems like regressions or tests that need updating, but I'm not sure which. Does it pass against the 3.15 api? You don't need to update the 3.15 apipie cache in this PR.

Fixed the tests and left some notes about what I found. I'd like to keep the 3.15 update here and cherry-pick it to the 0.21.z branch which would mean not squashing this PR on merge.

test/functional/activation_key/add_host_collection_test.rb
akofink: Yeah, seems like a regression to me. There are supported methods to test for this if we have to: `api_expects(...).at_most(2)`. I think they come from Mocha, but I'm not 100% sure https://mocha.jamesmead.org/Mocha/Expectation.html

chris1984:
@akofink did you need me to do anything next week when I get back from training to fix stuff?

jturel:
> So @jturel, I'm fine to merge this so long as we report the regressions wrt multiple API calls. Nice work figuring out the failures!

ack. lemme take a look around to see if I can figure out why those APIs are getting called twice before merging

akofink:
@chris1984 seems like jturel figured it out :) We should make sure tests on master continue to pass, but we can tackle that when you return.

jturel:
Found the problem in hammer-cli-foreman. Will clean up this PR once it's merged: https://github.com/theforeman/hammer-cli-foreman/pull/503

jturel:
@akofink it's green! :green_apple:

PR #727, opened 2020-03-12
Fixes #29310 - use environment_ids param for cv promote

jlsherrill:


jlsherrill:
difference in help output:
```
[vagrant@centos7-hammer-devel hammer-cli-katello]$ !diff
diff /tmp/output.old  /tmp/output.new -w
9,11d8
<  --environment-ids ENVIRONMENT_IDS                     Identifiers for Lifecycle Environment (--environment-ids is deprecated: Use --lifecycle-environment-ids instead)
<                                                        Comma separated list of values. Values containing comma should be quoted or escaped with backslash.
<                                                        JSON is acceptable and preferred way for complex parameters
16,18d12
<  --lifecycle-environment-ids LIFECYCLE_ENVIRONMENT_IDS Identifiers for Lifecycle Environment
<                                                        Comma separated list of values. Values containing comma should be quoted or escaped with backslash.
<                                                        JSON is acceptable and preferred way for complex parameters

```

PR #736, opened 2020-07-21
Fixes #30464 - Add description to repository info command

chris1984:
With patch:

```
[vagrant@hammer hammer-cli-katello]$ hammer repository info --id 1
Warning: An error occured while loading module hammer_cli_foreman_remote_execution.
ID:                 1
Name:               Repo 1
Label:              Repo_1
Description:        Updated description
```

Updated API doc as well because of an API change PR:

https://github.com/Katello/katello/pull/8848


jjeffers:
@chris1984 not sure if the investment is worth it for just that change, but it would be interesting to have a set of unit tests for the output fields in general. What do you think?

chris1984:
@jjeffers I am fine with making some unit tests, I will rework them and include them in this pr.

PR #739, opened 2020-08-28
Fixes #30735 - Add resolve traces command to Hammer

chris1984:
Requires Katello PR: https://github.com/Katello/katello/pull/8945

Output with patch:

```ruby
[vagrant@hammer ~]$ hammer host traces resolve --trace-ids [299]
[..................................................................................................................................................................................................................................................................................] [100%]1 task(s), 1 success, 0 fail
```

### DIFF
--- a/lib/hammer_cli_katello/host_traces.rb
+++ b/lib/hammer_cli_katello/host_traces.rb
@@ -14,6 +14,23 @@ module HammerCLIKatello
       end
       build_options
     end
+
+    class ResolveCommand < HammerCLIKatello::SingleResourceCommand
+      include HammerCLIForemanTasks::Async
+      resource :host_tracer, :resolve
+      command_name "resolve"
+
+      success_message _("Traces are being resolved with task %{id}.")
+      failure_message _("Could not resolve traces")
+
+      validate_options do
+        option(:option_trace_ids).required
+        option(:option_host_id).required
+      end
+
+      build_options
+    end
+
     autoload_subcommands
   end
 end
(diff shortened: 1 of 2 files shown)


jturel:
```
[vagrant@centos7-hammer-devel ~]$ hammer host traces resolve --help
Usage:
    hammer host traces resolve [OPTIONS]

Options:
 --async                       Do not wait for the task
 --host[-id]                   Set the current host context for the request. Name/Id can be used
 --trace-ids TRACE_IDS         Array of Trace Ids
                               Comma separated list of values. Values containing comma should be quoted or escaped with backslash.
                               JSON is acceptable and preferred way for complex parameters

```

```
[vagrant@centos7-hammer-devel ~]$ hammer host traces resolve --host-id=3 --trace-ids=1
Could not resolve traces:
  undefined method `job_invocation' for #<JobInvocation:0x00007f8be0e0c978>
```

I left a comment in the Katello PR on how to fix this

test/test_helper.rb
```diff
@@ -17,7 +17,7 @@
 require 'mocha/minitest'
 require 'hammer_cli'
 
-KATELLO_VERSION = Gem::Version.new(ENV['TEST_API_VERSION'] || '3.17')
+KATELLO_VERSION = Gem::Version.new(ENV['TEST_API_VERSION'] || '3.18')
```
jturel, line 20: Since this is going to be cherry-picked downstream I'm not sure if this will break anything since it's part of this PR? Maybe it's fine and only related to the tests. Ditto my comment for the above JSON

chris1984: I will make a separate pr to update the test data that way we are not updating it with this one

chris1984: Made https://github.com/Katello/hammer-cli-katello/pull/744 to separate out the test helper and new api data

jturel (approved):
ACK! tested in conjunction with the katello PR

PR #743, opened 2020-09-11
Fixes #30825 - Add back docker tag uploads with required params

sjha4:
To test:
1. Create a docker repository fedora/ssh for example and sync
2. From hammer, try adding a new tag.
hammer -d repository update --id $repo_id --organization-id $org_id --docker-tag some_name --docker-digest sha256:a6ecbb1553353a08936f50c275b010388ed1bd6d9d84743c7e8e7468e2acd82e

This change adds back the logic dealing with tagging manifests here, which I removed a while ago: https://github.com/Katello/hammer-cli-katello/pull/699/files#diff-e5e0e5d6b8dc7a2d91e2b54f479e2904L230 with couple of changes to maintain compatibility.

### DIFF
--- a/lib/hammer_cli_katello/repository.rb
+++ b/lib/hammer_cli_katello/repository.rb
@@ -249,6 +249,13 @@ module HammerCLIKatello
         if option(:option_product_name).exist?
           any(*organization_options).required
         end
+
+        if option(:option_docker_tag).exist? != option(:option_docker_digest).exist?
+          option(:option_docker_tag).rejected(
+            :msg => _('--docker-digest required with --docker-tag'))
+          option(:option_docker_digest).rejected(
+            :msg => _('--docker-tag required with --docker-digest'))
+        end
       end
 
       build_options(:without => [:unprotected]) do |o|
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/repository.rb
```diff
@@ -257,12 +264,60 @@ class UpdateCommand < HammerCLIKatello::UpdateCommand
...
+      rescue => e
+        @failure = true
+        logger.error e
+        output.print_error _("Failed to upload tag '%s' to repository.") % tag
+      ensure
+        content_upload_resource.call(:destroy, :repository_id => get_identifier, :id => upload_id)
+      end
+
+      def create_content_upload
+        response = content_upload_resource.call(:create, :repository_id => get_identifier, :size=>0)
```
sjha4, line 305: 1st change for compatibility: Adding the :size=>0 in case of tagging.

lib/hammer_cli_katello/repository.rb
```diff
@@ -257,12 +264,60 @@ class UpdateCommand < HammerCLIKatello::UpdateCommand
...
+      end
+
+      def import_uploads(uploads, opts = {})
+        publish_repository = opts.fetch(:last_file, false)
+        sync_capsule = opts.fetch(:last_file, false)
+        params = {:id => get_identifier,
+                  :uploads => uploads,
+                  publish_repository: publish_repository,
+                  sync_capsule: sync_capsule,
+                  content_type: "docker_tag"
```
sjha4, line 317: 2nd change for compatibility: Adding content_type param here to docker_tag because without this the default content type for docker, i.e manifest is assumed.
https://github.com/Katello/katello/blob/427975911bce0974413bc466683da5037b578ca4/app/lib/actions/katello/repository/import_upload.rb#L16

chris1984:
Starting to test

sjha4:
@chris1984 : Forgot to mention this only works for pulp2..The issues you were seeing earlier are all pulp3..Did not strike me at the time that defualt is pulp3 for docker in nightlies. :facepalm: 

To test:
vim ~/foreman/config/settings.plugins.d/katello.yaml
Un-comment  **:use_pulp_2_for_content_type:** and set
**:docker:                   true**

chris1984:
After making the change, and doing a restart of services, I created a new product and repo and was able to update a new tag. ACK

chris1984 (approved):

PR #746, opened 2020-09-26
Remove redundant text

swadeley:
Hello

I came here to fix the formatting; stray back tics in help output:

`Manipulate content credentials (i.e. GPG Keys)' 'on the server`

but as `i.e. GPG` is incorrect, because  content-credentials is more than just GPG keys (`e.g. GPG` would be OK), I thought to just remove it.

Thank you


chris1984 (approved):
ACK

swadeley:
Thank you.

PR #760, opened 2020-11-18
Fixes #31076 - hammer import  endpoint for cvv and library

parthaa:
This commit provides the bindings to import from a version to a view or
to the library.


parthaa:
Related to https://github.com/Katello/katello/pull/9042

parthaa:
Follow the steps listed here https://github.com/Katello/katello/pull/9042#issuecomment-729947436

chris1984:
Starting to test

chris1984 (approved):
Looks amazing, thanks @parthaa

lib/hammer_cli_katello/content_view.rb
```diff
@@ -106,10 +106,12 @@ class CreateCommand < HammerCLIKatello::CreateCommand
       failure_message _("Could not create the content view")
 
       option ["--composite"], :flag, _("Create a composite content view")
-
+      option ["--import-only"], :flag, _("Designate this Content View for "\
```
jturel, line 109: Does this undo the server-side mechanism we use to conditionally show or hide the parameter?

lib/hammer_cli_katello/content_import.rb
```diff
@@ -0,0 +1,61 @@
...
+      desc _('Imports archive to a content view version')
+      action :version
+      command_name "version"
+
+      include HammerCLIForemanTasks::Async
+      include ContentImportCommon
+    end
+
+    class LibraryCommand < HammerCLIKatello::SingleResourceCommand
+      desc _('Imports archive to the organization library lifecycle-environment')
```
jturel, line 51: suggestions

1. Imports a content archive to an organization's library
2. Imports a content archive to an organization's library lifecycle environment (no hyphen)

jturel (commented):
LGTM. minor feedback

lib/hammer_cli_katello/content_view.rb
parthaa: No, we do the same for composite as you see above. So trying to stay consistent with other boolean params

chris1984 (approved):
ACK, was able to get the export to show the path in quotes and metadata file generated during an export, and was able to import both library and version with the updated changed. Thanks @parthaa

PR #765, opened 2020-12-01
Fixes #29732 - correct letter case of fields

upadhyeammit:


chris1984 (commented):
@upadhyeammit `UUID` and `ID` in the webui are all capitol, don't we want to keep it the same across the CLI as well?

upadhyeammit:
> @upadhyeammit `UUID` and `ID` in the webui are all capitol, don't we want to keep it the same across the CLI as well?

The output stays same with this change,

~~~
---|----------------------------------|------|----------|----------|---------|---------|---------------------|---------------------|-----------|---------
ID | UUID                             | NAME | TYPE     | CONTRACT | ACCOUNT | SUPPORT | START DATE          | END DATE            | QUANTITY  | CONSUMED
---|----------------------------------|------|----------|----------|---------|---------|---------------------|---------------------|-----------|---------
1  | 4028fad275fe64620175fe7111c20003 | test | Physical |          |         |         | 2020/11/25 08:07:04 | 2049/12/01 00:00:00 | Unlimited | 0       
---|----------------------------------|------|----------|----------|---------|---------|---------------------|---------------------|-----------|---------
~~~

Change is in predefined field sets, if you keep it as ID it does not work,

~~~
Predefined field sets:
  -----------|-----|--------
  FIELDS     | ALL | DEFAULT
  -----------|-----|--------
  Id         | x   | x
  Uuid       | x   | x
~~~

I can see same capitalizaiton is being followed by hammer-cli-foreman(Id),
~~~
---|-------------------|------------------|------------|----|-------------------|---------------|--------------|----------------------
ID | NAME              | OPERATING SYSTEM | HOST GROUP | IP | MAC               | GLOBAL STATUS | CONTENT VIEW | LIFECYCLE ENVIRONMENT
---|-------------------|------------------|------------|----|-------------------|---------------|--------------|----------------------
1  | dina-navejar.test | centos 7.0       |            |    | 52:54:00:29:08:f7 | Error         |              |                      
---|-------------------|------------------|------------|----|-------------------|---------------|--------------|----------------------

Predefined field sets:
  -----------------------|-----|---------|-----
  FIELDS                 | ALL | DEFAULT | THIN
  -----------------------|-----|---------|-----
  Id                     | x   | x       | x
~~~

chris1984 (approved):
That makes sense then, looks good. ACK

PR #766, opened 2020-12-02
Refs #31333 - Add assert single commit action

chris1984:
This PR adds a non-blocking check to assert a single commit like we have in Katello:

https://github.com/Katello/katello/blob/master/.github/workflows/pull_request.yml

Screenshot of it working:

https://nimbusweb.me/s/share/4938240/abcdg9p85tb43fs35vfm


jeremylenz (approved):
ACK

Now you should write a Chrome extension that disables the "Rebase and Merge" button when this check is failed ;)

PR #767, opened 2020-12-02
Fixes #31438 - Add generate-metadata command for content-export

jturel:


parthaa:
I like the ```--task-id``` option. Just make sure it errors out gracefully if the task is not completed or export_history_id is not available.

lib/hammer_cli_katello/content_export_helper.rb
```diff
@@ -1,12 +1,16 @@
 module HammerCLIKatello
   module ContentExportHelper
     include ApipieHelper
+
     def execute
       response = super
       if option_async? || response != HammerCLI::EX_OK
```
parthaa, line 7: prefer a split that looks like
``` ruby
if option_async?
    output.print_message(......)
      HammerCLI::EX_OK
elsif response != HammerCLI::EX_OK
    reponse
else
    .....
```

lib/hammer_cli_katello/content_export_helper.rb
```diff
@@ -17,6 +21,10 @@ def execute
       end
     end
 
+    def send_request
```
parthaa, line 24: shouldn't be necessary. we are overriding task_progress  itself to make it store ```@task``` so no need for the send_request to do the same

jturel: I don't think task_progress is called from generate-metadata

parthaa: still not needed I think... Lemme test with it and without and tell ya :).

lib/hammer_cli_katello/content_export.rb
```diff
@@ -6,6 +6,44 @@ class ContentExport < HammerCLIKatello::Command
...
+      command_name 'generate-metadata'
+
+      include ContentExportHelper
+
+      option "--task-id", "TASK_ID",
+             _("Generate metadata based on output of the specified export task"),
+             :attribute_name => :option_task_id,
+             :required => false
+
+      option "--export-history-id", "EXPORT_HISTORY_ID",
```
parthaa, line 23:
```suggestion
      option "--id", "EXPORT_HISTORY_ID",
```

lib/hammer_cli_katello/content_export_helper.rb
```diff
@@ -1,12 +1,18 @@
...
   module ContentExportHelper
     include ApipieHelper
+
     def execute
       response = super
-      if option_async? || response != HammerCLI::EX_OK
+      if option_async?
+        output.print_message _("Once the task completes the export metadata must be generated "\
+          + "with the command: hammer content-export generate-metadata "\
+          + "--task-id %s") % @task['id']
```
parthaa, line 10:
```suggestion
          + "--task-id %{id}")
```

parthaa: If you made that change you can delete send_request

lib/hammer_cli_katello/content_export_helper.rb
parthaa: Ok I am wrong.. Remove the task_progress and keep this guy intact

lib/hammer_cli_katello/content_export_helper.rb
```diff
@@ -1,12 +1,18 @@
 module HammerCLIKatello
   module ContentExportHelper
     include ApipieHelper
+
     def execute
       response = super
-      if option_async? || response != HammerCLI::EX_OK
+      if option_async?
+        output.print_message _("Once the task completes the export metadata must be generated "\
+          + "with the command: hammer content-export generate-metadata "\
```
parthaa, line 9: Can ya add quotes to the command or a new line. Like

```suggestion
          + "with the command:\n hammer content-export generate-metadata "\
```

parthaa:
ACK once the comments are addressed :)

lib/hammer_cli_katello/content_export_helper.rb
parthaa: or even
```ruby
        output.print_message _("Once the task completes the export metadata must be generated "\
          + "with the command:")
        output.print_message("hammer content-export generate-metadata --task-id #{@task['id']}")
```
I dont think the command itself needs translation right?

chris1984:
@parthaa thanks for reviewing, a quick peek, looks good other than the comments you left.

jturel:
GH not understanding that i've pushed changes here ...

PR #782, opened 2021-02-02
Refs #30939 - Remove Puppet commands and Katello 4.0 removals.

chris1984:
With Katello 4.0 we are removing all Puppet items. This PR also takes care of removing Pulp2 from the ping output as well as the deprecations being removed in the Katelllo PR linked to the Redmine issue.


jjeffers (commented):
LGTM

PR #792, opened 2021-03-15
Drop hammer_cli_foreman_docker

ehelms:
The foreman_docker plugin has been discontinued and it's Hammer
plugin.

### DIFF
--- a/hammer_cli_katello.gemspec
+++ b/hammer_cli_katello.gemspec
@@ -50,7 +50,6 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency 'hammer_cli_foreman'
   gem.add_dependency 'hammer_cli_foreman_tasks'
-  gem.add_dependency 'hammer_cli_foreman_docker'
 
   gem.add_development_dependency 'rake'
   gem.add_development_dependency 'thor'


chris1984 (approved):
ACK, @ehelms can you rebase?

ehelms:
Rebased.

Packaging PR for this and the hammer_cli_bootdisk change: https://github.com/theforeman/foreman-packaging/pull/6441

ehelms:
Looks like this was not as "simple" (or even obvious) of a change. The hammer_cli_foreman_docker project appears to have contained commands that are only related to Katello, which is not obvious:

https://github.com/theforeman/hammer_cli_foreman_docker/blob/master/lib/hammer_cli_foreman_docker/docker_manifest.rb
https://github.com/theforeman/hammer_cli_foreman_docker/blob/master/lib/hammer_cli_foreman_docker/docker_tag.rb

You can see that we test for them, e.g. https://github.com/theforeman/forklift/blob/master/bats/fb-katello-content.bats#L312-L317

So the question I have is, do we need to keep these commands going forward? And if so, should I just move these files into hammer-cli-katello directly?

@chris1984 @jlsherrill

jlsherrill:
I think it makes sense to move those commands to hammer-cli-katello

PR #793, opened 2021-03-15
Fixes #32102: Drop requirement on hammer_cli_bootdisk

ehelms:
There is no code requirements on bootdisk, and while this functionality
might be useful to a Katello user, dropping this can loosen coupling
to other plugins and give Katello more flexibility.


chris1984 (approved):
ACK

PR #797, opened 2021-03-31
Refs #32241 - Auto create cv on import

parthaa:


chris1984 (approved):
I was able to have the cv created for me with only just making the repos/product:

```bash
$ hammer content-import version --organization=import --path /var/lib/pulp/exports/Default_Organization/view/1.0/2021-04-05T15-51-44-04-00
[...............................................................................................................................] [100%]
```

PR #798, opened 2021-04-06
Fixes #32241 - Updated unit tests for auto create cv

parthaa:
The unit tests and the apipie json were not updated when the auto create
cv PR got merged. This commit fixes that issue


chris1984 (approved):
ACK

PR #809, opened 2021-06-23
Fixes #32868 - Make --unlimited-hosts a flag so it doesn't require a value

jeremylenz:
Previously:

#### `hammer activation-key create`: 
* takes unlimited-hosts argument
* errors if you pass a value such as `--unlimited-hosts=yes`
* Behaves correctly.

#### `hammer host-collection create`:
* takes unlimited-hosts argument
* errors if you do NOT pass a value
* Behaves incorrectly.

With this patch, `hammer host-collection create` now takes `--unlimited-hosts` as a flag rather than a parameter, meaning that it behaves the same way as `hammer activation-key create`. Thus, hammer's behavior is more consistent.

### DIFF
--- a/lib/hammer_cli_katello/host_collection.rb
+++ b/lib/hammer_cli_katello/host_collection.rb
@@ -52,6 +52,10 @@ module HammerCLIKatello
 
       option "--unlimited-hosts", :flag, "Set hosts max to unlimited"
 
+      validate_options :before, 'IdResolution' do
+        all(:option_unlimited_hosts, :option_max_hosts).rejected
+      end
+
       build_options
     end
 
--- a/lib/hammer_cli_katello/host_collection.rb
+++ b/lib/hammer_cli_katello/host_collection.rb
@@ -50,6 +50,8 @@ module HammerCLIKatello
       success_message _("Host collection created.")
       failure_message _("Could not create the host collection")
 
+      option "--unlimited-hosts", :flag, "Set hosts max to unlimited"
+
       build_options
     end
 
--- a/test/functional/host_collection/create_test.rb
+++ b/test/functional/host_collection/create_test.rb
@@ -29,6 +29,17 @@ module HammerCLIKatello
       run_cmd(%w(host-collection create --name hc1 --organization org1))
     end
 
+    it 'allows unlimited-hosts flag with no arguments' do
+      api_expects(:organizations, :index) { |par| par[:search] == "name = \"org1\"" }
+        .returns(index_response([{'id' => 1}]))
+
+      api_expects(:host_collections, :create) do |par|
+        par['unlimited_hosts'] == true
+      end
+
+      run_cmd(%w(host-collection create --name hc1 --organization org1 --unlimited-hosts))
+    end
+
     it 'allows organization label' do
       api_expects(:organizations, :index) { |par| par[:search] == "label = \"org1\"" }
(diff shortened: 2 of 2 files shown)


jeremylenz:
behavior before patch:
```
[vagrant@centos7-katello-devel hammer-cli-katello]$ hammer host-collection create --unlimited-hosts
Could not create the host collection:
  Error: Option '--unlimited-hosts': Value must be one of true/false, yes/no, 1/0..
  
  See: 'hammer host-collection create --help'.
[vagrant@centos7-katello-devel hammer-cli-katello]$ hammer activation-key create --unlimited-hosts=yes
Could not create the activation key:
  Error: Too many arguments.
  
  See: 'hammer activation-key create --help'.
```

lib/hammer_cli_katello/host_collection.rb
```diff
@@ -50,6 +50,8 @@ def request_params
       success_message _("Host collection created.")
       failure_message _("Could not create the host collection")
 
+      option "--unlimited-hosts", :flag, "Set hosts max to unlimited"
```
parthaa, line 53:
```suggestion
      option "--unlimited-hosts", :flag, "Set hosts max to unlimited"
      
      validate_options :before, 'IdResolution' do
        all(:option_unlimited_hosts, :option_max_hosts).rejected
      end
```
We don't want the user to do something like -> 
```
$ hammer host-collection create --organization-id=1 --unlimited-hosts --name=great2 --max-hosts=1
```

parthaa:
Mostly works well other than that one comment

test/functional/host_collection/create_test.rb
```diff
@@ -29,6 +29,17 @@ module HammerCLIKatello
       run_cmd(%w(host-collection create --name hc1 --organization org1))
     end
 
+    it 'allows unlimited-hosts flag with no arguments' do
+      api_expects(:organizations, :index) { |par| par[:search] == "name = \"org1\"" }
+        .returns(index_response([{'id' => 1}]))
+
+      api_expects(:host_collections, :create) do |par|
+        par['unlimited_hosts'] == true
```
parthaa, line 37:
```suggestion
        assert par['unlimited_hosts']
```

parthaa: Interesting.. I thought you always wanted to assert looking at some of the test .. But then I see other tests doing what you are doing.

parthaa: Ok I take back. I tested with
```
 par['unlimited_hosts'] == false
```
and it failed correctly.

parthaa: sounds like either work

jeremylenz:
@parthaa committed your suggestion. It's on the dev box I had to destroy, so I didn't test it. But based on the same code in activation_keys I think it's fine :)

parthaa:
> @parthaa committed your suggestion. It's on the dev box I had to destroy, so I didn't test it. But based on the same code in activation_keys I think it's fine :)

works well
```
$ hammer  host-collection create --organization-id=1 --unlimited-hosts --name=great3 --max-hosts=1
Could not create the host collection:
  Error: You can't set all options --unlimited-hosts, --max-hosts at one time.
  
  See: 'hammer host-collection create --help'.
```

parthaa (approved):

PR #812, opened 2021-08-03
Refs #33182 - hide broken options --ids

lfu:


chris1984 (approved):
ACK works great, thanks @lfu 

output with the patch:

```bash
[vagrant@hammer hammer-cli-katello]$ hammer host errata apply -h
Usage:
    hammer host errata apply [OPTIONS]

Options:
 --async                       Do not wait for the task
 --errata-ids ERRATA_IDS       List of Errata ids to install. Will be removed in a future release
                               Comma separated list of values. Values containing comma should be quoted or escaped with backslash.
                               JSON is acceptable and preferred way for complex parameters
 --host[-id]                   Host name/id
 --search SEARCH               Search string for erratum to perform an action on
```

PR #815, opened 2021-08-31
Fixes #33381 - Add back Puppet search to Foreman

chris1984:


ofedoren (approved):
Thanks, @chris1984! Tested with 2.5-stable branches and it worked just fine. Also, tests are passing localy with those branches, which is right.

P.S. Tests are failing here since CI uses the latest branches of hammer-cli[-foreman] (I've fixed h-c-katello master branch already).

chris1984:
> Thanks, @chris1984! Tested with 2.5-stable branches and it worked just fine. Also, tests are passing localy with those branches, which is right.
> 
> P.S. Tests are failing here since CI uses the latest branches of hammer-cli[-foreman] (I've fixed h-c-katello master branch already).

I will make a pr to use the correct version of hammer-cli[-foreman] for the CI

PR #817, opened 2021-10-06
Fixes #33327 - host errata recalculate does not need --async flag

lfu:
Hammer host errata recalculate calls `/hosts/:host_id/errata/applicability` which puts a `generate_host_applicability` event into the queue.

Option `--async` is no longer necessary for `/hosts/:host_id/errata/applicability`.

`/hosts/:host_id/errata/applicability` does not return back a [task](https://github.com/Katello/katello/blob/c35f3100fda7fcdd89b34e8324b3e82c4b8b491e/app/controllers/katello/api/v2/host_errata_controller.rb#L77).

**To Test**
- sync https://partha.fedorapeople.org/test-repos/multi-errata/ 
- Point a host to this repo and yum install armadillo-0.2-1 
- hammer host errata recalculate --host-id <host_id>

**Before**
`Error: Please mention appropriate attribute value`
**After**
`Errata recalculated.`

### DIFF
--- a/lib/hammer_cli_katello/host_errata.rb
+++ b/lib/hammer_cli_katello/host_errata.rb
@@ -46,10 +46,9 @@ module HammerCLIKatello
     end
 
     class RecalculateCommand < HammerCLIKatello::SingleResourceCommand
-      include HammerCLIForemanTasks::Async
       resource :host_errata, :applicability
       command_name "recalculate"
-      success_message _("Errata recalculated with task %{id}.")
+      success_message _("Errata recalculation started.")
       failure_message _("Could not recalculate errata")
 
       build_options
--- a/test/functional/host/errata/recalculate_test.rb
+++ b/test/functional/host/errata/recalculate_test.rb
@@ -8,33 +8,19 @@ describe 'recalculate errata' do
   end
 
   let(:host_id) { 1 }
-  let(:task_id) { '5' }
-  let(:response) do
-    {
-      'id' => task_id,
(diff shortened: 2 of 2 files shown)


lib/hammer_cli_katello/host_errata.rb
```diff
@@ -46,10 +46,9 @@ class InfoCommand < HammerCLIKatello::ErratumInfoCommand
     end
 
     class RecalculateCommand < HammerCLIKatello::SingleResourceCommand
-      include HammerCLIForemanTasks::Async
       resource :host_errata, :applicability
       command_name "recalculate"
-      success_message _("Errata recalculated with task %{id}.")
+      success_message _("Errata recalculated.")
```
jeremylenz, line 51: I think different wording may be appropriate here.  This is a past-tense message, while in reality the recalculation may still be in progress when the user sees this.  Perhaps something a bit more accurate like
```suggestion
      success_message _("Errata recalculation started.")
```
cc @jturel to get his thoughts as well

jeremylenz: seems like calling `success_message` is the standard way to do this, based on its 104 existing uses in this repo I'm okay with this.

chris1984: +1 for eventually getting sucess messages from the server, but def out of the scope of this pr to remove the 104 use cases.

lfu: Message updated.

PR #822, opened 2021-11-01
Refs #33292 - updates content upload for ostree content

jjeffers:
This PR adjusts uploads to handle inputs for ostree content. It adds the repository name for OStree content ("refs") as an option due to the way the request from hammer bridges 2 distinct requests to Katello (content upload and content import).


chris1984 (approved):
ACK

PR #823, opened 2021-11-05
test fixing

chris1984:


test/functional/content_view/version/republish_repositories_test.rb
```diff
@@ -26,7 +26,7 @@ module HammerCLIKatello
       expect_foreman_task('3')
 
       result = run_cmd(%w(content-view version republish-repositories --organization org1
-                          --content-view cv --version 1))
+                          --content-view cv --version 1 --force true))
```
ofedoren, line 29: Isn't it dangerous that few checks wont apply? I don't know what this flag is for exactly, but doesn't it mean that if this flag is set then some logic in the command is dropped?

test/functional/package/list_test.rb
```diff
@@ -72,23 +72,15 @@ module HammerCLIKatello
     describe 'environment options' do
       it 'may be specified environment name' do
         org_id = 2
-        env_name = "dev"
+        env_id = 3
         expected_id = 6
-        expect_lifecycle_environment_search(org_id, env_name, expected_id)
+        expect_lifecycle_environment_search(org_id, env_id, expected_id)
```
ofedoren, line 77: If it searches for environment why you drop the name here? I assume that it does `index` call with the name to grab the id.

test/functional/capsule/content/remove_lifecycle_environment_test.rb
```diff
@@ -15,14 +16,14 @@ class Content
...
-          run_cmd(%w(capsule content remove-lifecycle-environment --id 1 --environment-id 2))
+          run_cmd(%w(capsule content remove-lifecycle-environment --id 1 --lifecycle-environment-id 2))
         end
 
         it 'resolves environment ID from name' do
           expect_lifecycle_environment_search('3', 'env2', 2)
           expect_lifecycle_environment_remove('id' => 1, 'environment_id' => 2)
 
-          run_cmd(%w(capsule content remove-lifecycle-environment --id 1 --environment env2
+          run_cmd(%w(capsule content remove-lifecycle-environment --id 1 --lifecycle-environment env2
```
ofedoren, line 26: If you're dropping support for `--environment` in tests, maybe it's time to actually drop it from the commands?

chris1984: @ofedoren I was just trying to fix the tests, do you think we should keep it?

test/functional/package/list_test.rb
chris1984: ok fair point I will switch it back

test/functional/content_view/version/republish_repositories_test.rb
chris1984: Sadly, this command now won't work unless --force is passed in. When I try it without it, it companies it needs and when i set it to false, it says it must be true. This is the pr that introduced it.

https://github.com/Katello/katello/commit/79e606ac3ffcb416c3c03baeca12e083e24bbe84

ofedoren: Ah, right. Well, if it *is* required in the API in such way, then it's OK, my bad :/

test/functional/module_stream/info_test.rb
```diff
@@ -11,11 +11,11 @@ module HammerCLIKatello
 
     it 'resolves ID from name and repo id' do
       api_expects(:module_streams, :index)
-        .with_params(search: "name = \"duck\"", name: "duck", repository_id: "1")
-        .returns(index_response([{'id' => 1}]))
+        .with_params(search: "name = \"duck\"", name: "duck", repository_id: 1)
```
parthaa, line 14:
```suggestion
        .with_params(search: "name = \"duck\"", name: "duck", repository_id: "1")
```

PR #826, opened 2021-11-11
Fixes #33837 - Add Generic Content Unit Support

rverdile:
This adds support for the list/info commands with generic content units and currently works for python packages and ostree refs. It depends on this Katello change: https://github.com/Katello/katello/pull/9776

**I could use feedback on handling the output for the list and info commands:**

- For the list command, I overrode the print_data method and built the fields dynamically because ostree refs don't have filename metadata and I didn't want to display an empty column. I couldn't find a hammer-friendly way to leave out nil columns. 
 
- For the info command, I overrode the print_data method and manually created the output. This was to better handle printing the "additional_metadata" contents, which is a hash of varying metadata.

Edit: 
- I removed additional_metadata for now so there's no hacky way of printing. 
- I don't think I'm using it incorrectly, but `hide_blank` isn't working for ListCommand.


ofedoren:
Please, don't use such approach for the output unless you really have to do so. If there is a lack of functionallity you'd gladly have, please feel free to implement that in `hammer-cli` or `hammer-cli-foreman` :) If you need assistance, please reach me and describe what would help you and we can implement that in hammer core :)

> * For the list command, I overrode the print_data method and built the fields dynamically because ostree refs don't have filename metadata and I didn't want to display an empty column. I couldn't find a hammer-friendly way to leave out nil columns.
Well, have you tried this: `field :bla, Fields::Field, hide_missing: true, hide_blank: true` ? `hide_missing` should hide the field is not present in the response and `hide_blank` should hide if the field is present but has `nil` as a value if set to `true`.
> * For the info command, I overrode the print_data method and manually created the output. This was to better handle printing the "additional_metadata" contents, which is a hash of varying metadata.
> 
Can I have a machine to check what's the response is and how it behaves? Or if it is possible, could you provide the API response you want to show/to deal with?

rverdile:
@ofedoren Thanks!

If you want to spin up a box with ostree and python types, you can follow these steps to spin up a centos8 stream dev box. And I'm happy to provide steps to sync content for these types if needed. https://github.com/theforeman/forklift/blob/master/docs/development.md#deploy-a-katello-development-environment.

If you don't want to go that route, here is what a single content unit looks like in the api:

**ostree_ref** 
{
      "id": 2,
      "name": "rawhide",
      "filename": null,
      "pulp_id": "/pulp/api/v3/content/ostree/refs/62790095-a258-489d-a26b-7e231df029e6/",
      "version": null,
      "content_type": "ostree_ref",
      "additional_metadata": null
}

**python_package**
{
      "id": 2717,
      "name": "pulp-python",
      "filename": "pulp_python-3.0.0-py3-none-any.whl",
      "pulp_id": "/pulp/api/v3/content/python/packages/dd0f6533-edb3-4114-b39b-97f4f2dab89a/",
      "version": "3.0.0",
      "content_type": "python_package",
      "additional_metadata": {
        "sha256": "80854f507e6b9500dfb945cd331c12151a2fd583fb3d97230fc6224caafbf025",
        "package_type": "bdist_wheel"
      }
    },

**For the list command**
I completely agree with your approach for the List command. I had not found that `hide_blank` option before. I made that change, but it's still showing an empty column for me. They may just require some debugging on my part, however.

**For the Info command,** 
You can see above the "additional_metadata" field contains a hash. So in this case, I want Info to show `sha256` and `package type` for a python package. I don't want the user to see "additional_metadata". And these fields in "additional_metadata" can vary; they're not known ahead of time.

Is there a more hammer-friendly way to do this? Or if there's a change I could make in a different hammer repo, could you point in the right direction?

chris1984 (approved):
ACK, this looks good to me, holding off on merging until @ofedoren gets back because he started the review.

jlsherrill:
@rverdile removed the complicated part and we're gonna worry about it later.  What's here is pretty straight forward,  i propose we go ahead and merge.

PR #830, opened 2021-12-15
Fixes #34156 - Set async param for import uploads with ostree refs

rverdile:
OSTree refs take a while to upload, so Foreman Tasks will timeout unless the async param is set. This sets the async param when ostree_ref is the content type and adds a "success" message.

### DIFF
--- a/lib/hammer_cli_katello/repository.rb
+++ b/lib/hammer_cli_katello/repository.rb
@@ -389,6 +389,8 @@ module HammerCLIKatello
         o.expand.including(:products, :organizations)
       end
 
+      option "--async", :flag, _("Do not wait for the task.")
+
       option "--ostree-repository-name", "OSTREE REPOSITORY NAME",
         _("Name of OSTree repository in archive."),
         :attribute_name => :option_ostree_repository_name,
@@ -415,6 +417,7 @@ module HammerCLIKatello
           update_content_upload(upload_id, repo_id, file)
         end
         opts[:ostree_repository_name] = options["option_ostree_repository_name"]
+        opts[:filename] = filename
         results = import_uploads([
           {
             id: upload_id,
@@ -423,7 +426,7 @@ module HammerCLIKatello
             size: file.size,
             checksum: checksum
           }], opts)
-        print_results(filename, results)
+        print_results(filename, results, options["option_async"])
       ensure
         if upload_id
           content_upload_resource.call(:destroy, :repository_id => get_identifier, :id => upload_id)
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/repository.rb
```diff
@@ -478,10 +480,14 @@ def import_uploads(uploads, opts = {})
...
 
       def print_results(name, results)
-        if results.empty? || results.dig('output', 'upload_results').empty?
+        if results.empty? || results.dig('output', 'upload_results')&.empty?
           print_message _("Successfully uploaded file '%{name}'") % {
             :name => name
           }
+        elsif results["pending"]
+          print_message _("Started uploading file '%{name}'") % {
+            :name => name
```
jlsherrill, line 489: I think we should poll the task to completion like we do with other tasks:    https://github.com/Katello/hammer-cli-katello/blob/master/lib/hammer_cli_katello/content_view.rb#L190

They use that HammerCLIForemanTasks::Async module to do all the polling for us.  (I'm not sure how hard that is to use here)

lib/hammer_cli_katello/repository.rb
```diff
@@ -465,10 +465,12 @@ def update_content_upload(upload_id, repo_id, file)
       def import_uploads(uploads, opts = {})
         publish_repository = opts.fetch(:last_file, false)
         sync_capsule = opts.fetch(:last_file, false)
+        async = options["option_content_type"] == 'ostree_ref'
```
jlsherrill, line 468: I think all types should use the async option IMO.  There are other ways for the task to take too long (such as if all pulp workers are busy syncing repos).   That might make it easier below if we want this command to poll tasks always

rverdile:
This doesn't work correctly yet. I pushed for safe-keeping.

rverdile:
Because `UploadContentCommand` overrides `def execute`, the Async module seems to have no affect when used. However, the command will already exit the hammer process when the `import_uploads` API returns a task, which will be the in-progress task, since the async API param is set.

Also, because `ContentUploadsController` does not create a task, the hammer process will hang while those API calls are made (which can take a while).

chris1984:
@rverdile @jlsherrill is this ok to test, or still work in progress?

rverdile:
@chris1984 still a work in progress

rverdile:
I added the `--async` flag. Setting it will return a task ID for the upload and close the hammer process. Without it, the hammer process will continue running and there will be a progress bar showing the progress of the upload.

rverdile:
@jlsherrill is this what you were looking for with async?

rverdile:
I updated the tests so that they work with `task_progress`. `print_results` also got a lot more complicated because I wanted to keep the same output as before (when not using async flag).

chris1984:
@rverdile is this good for a merge? It looks good as far as code goes and @jjeffers tested it

chris1984 (approved):

rverdile:
@chris1984 yup this is good to merge

PR #831, opened 2021-12-16
Fixes #34174 - Add content view ID to content-view component list

rverdile:
Test by creating content views and a composite content view. List the component content views with:

`hammer content-view component list --composite-content-view-id <id>`

### DIFF
--- a/lib/hammer_cli_katello/content_view_component.rb
+++ b/lib/hammer_cli_katello/content_view_component.rb
@@ -84,16 +84,16 @@ module HammerCLIKatello
       include OrganizationOptions
       include CompositeContentViewNameResolvable
       output do
-        field :id, _("Id")
+        field :content_view_id, _("Content View Id")
         field :content_view_name, _("Name")
         field :version, _("Version")
+        field :id, _("Component Id")
         field :current_version, _("Current Version")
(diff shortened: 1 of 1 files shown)


lib/hammer_cli_katello/content_view_component.rb
```diff
@@ -84,16 +84,16 @@ class ListCommand < HammerCLIKatello::ListCommand
       include OrganizationOptions
       include CompositeContentViewNameResolvable
       output do
-        field :id, _("Id")
+        field :id, _("Component Id")
         field :content_view_name, _("Name")
         field :version, _("Version")
+        field :content_view_id, _("Id")
```
sjha4, line 90: This works well. We should probably switch the order to show the content view Id first and component Id in this column instead..Thoughts?

rverdile: Sounds good. Should it also read "Content View Id" instead of "Id"?

sjha4: That would make it clearer too.. 👍🏼

rverdile:
updated

sjha4 (approved):

chris1984:
ACK

PR #840, opened 2022-03-11
fixes #34594 - Mirror Policy shows up in repo info

parthaa:
Mirroring policy is displayed instead of the defunct mirror on sync.

```bash
$ hammer  repository info --id=2|grep Mirror
Mirroring Policy:        Complete Mirroring
```


chris1984 (approved):
ACK

PR #843, opened 2022-04-19
Fixes #34762 - Add repo republish to Hammer

chris1984:
Steps to test:

* Apply patch to your hammer devel box or hammer location
* Try new hammer command out `hammer repository republish --id x --force true"
* See if you see the dynflow task kick off on your katello devel box

### DIFF
--- a/lib/hammer_cli_katello/repository.rb
+++ b/lib/hammer_cli_katello/repository.rb
@@ -567,6 +567,33 @@ module HammerCLIKatello
       end
     end
 
+    class RepublishCommand < HammerCLIKatello::SingleResourceCommand
+      extend RepositoryScopedToProduct
+      include HammerCLIForemanTasks::Async
+      include OrganizationOptions
+
+      validate_repo_name_requires_product_options
+      action :republish
+      command_name "republish"
+      desc _("Forces a republish of the specified repository.")
+
+      success_message _("Repository republished.")
+      failure_message _("Could not republish the repository.")
+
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/repository.rb
```diff
@@ -567,6 +567,17 @@ class RemoveContentCommand < HammerCLIKatello::SingleResourceCommand
       end
     end
 
+    class RepublishCommand < HammerCLIKatello::SingleResourceCommand
+      action :republish
```
sjha4, line 571: Should we include HammerCLIForemanTasks::Async to this as this could take a while to finish for larger repos?

sjha4: Also extend RepositoryScopedToProduct in this?

chris1984: @sjha4 updated

sjha4: We'll need similar params and validations like the other commands in the file to allow repository name to be scoped to product and org etc? See reclaim-space command for example.. 
include OrganizationOptions
      validate_repo_name_requires_product_options

chris1984: @sjha4 updated

sjha4 (approved):
Can run the hammer repository republish command with repo-id/ repo-name + product + org..
Task progress appears on the output as expected and the metadata generate task is started on the server against the correct repository.
Ack 👍🏼

PR #851, opened 2022-06-07
Fixes #35032 - Add content-view override to ak info command

chris1984:
* Apply patch to hammer devel box or in the hammer directory on your katello server
* In the UI create an activation key, attach a subscription, and disable or enable (override) a repo
* Run the command from hammer `hammer activation-key info --id 1`
* Verify it looks like the output below:

```
Name:                  foo
Id:                    1
Host Limit:            Unlimited
Auto Attach:           true
Lifecycle Environment: Library
Content View:          Default Organization View
Host Collections:

Content Overrides:
 1) Content Label: Default_Organization_Custom_Zoo
    Name:          enabled
    Value:         0
System Purpose:
    Service Level:
    Purpose Usage:
    Purpose Role:
    Purpose Addons:
```

### DIFF
--- a/lib/hammer_cli_katello/activation_key.rb
+++ b/lib/hammer_cli_katello/activation_key.rb
@@ -54,10 +54,10 @@ module HammerCLIKatello
       output do
         field :name, _("Name")
         field :id, _("Id")
-        field :description, _("Description")
+        field :description, _("Description"), Fields::Field, :hide_blank => true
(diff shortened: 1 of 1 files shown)


lfu:
Before
Screenshot from 2022-06-08 14-37-22


After
Screenshot from 2022-06-08 14-39-05

lfu (approved):
LGTM 👍🏻

PR #860, opened 2022-07-30
Refs #35296 - New Custom CDN type

parthaa:


parthaa:
Goes with -> https://github.com/Katello/katello/pull/10217

PR #863, opened 2022-08-29
Refs #35442 - Properly vets repository exports

parthaa:
Connected to https://github.com/Katello/katello/pull/10254

Try the steps listed here -> https://github.com/Katello/katello/pull/10254#issue-1353714691


chris1984 (approved):
ACK works well, only comment I have is when I exported I did not see the

`Unable to export repository 'Ansible' because it does not have an exportable content type.`

All I got back was:

```bash
[root@centos8-stream-katello-nightly ~]# hammer content-export complete repository --id=4 --format=syncable
Could not export the repository:
  500 Internal Server Error
```

Is there a way to have hammer show the error message instead of a 500? If not i'm happy to still merge this, i just think for a better user experience we should show why.

PR #869, opened 2022-11-02
Fixes #35666 - hammer repository types command is missing creatable

chris1984:
What are the changes introduced in this pull request?
* Added `build_options` to the class method so it can pull in all the params

What are the testing steps for this pull request?
* Run `hammer hammer repository types -h and see if the creatable param is now listed'


lfu (approved):
LGTM 👍🏻

PR #870, opened 2022-11-02
Remove Ruby 2.5 in unit tests

chris1984:
* Ruby 2.5 is not longer being used in Katello/Satellite since Satellite 6.8
* Ruby 2.5 unit tests started failing with:
```ruby
/home/runner/work/hammer-cli-katello/hammer-cli-katello/vendor/bundle/ruby/2.5.0/gems/mocha-2.0.0/lib/mocha/mock.rb:1:in `require': cannot load such file -- ruby2_keywords (LoadError)
```
* Fixed the failing 2.7 unit test as well in thi spr

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


chris1984:
@lfu updated to include the test fix here

PR #873, opened 2022-11-14
Fixes #35750 - ignore pagination when looking up cvv for cv purge

chris1984:
* Sent off per page of `100` to get all the content view versions in our collection call

* Create a content view and make 5 versions
* Change `Entries per page` to 2
* Apply PR
* Try to purge the content view, before the patch it would say, no versions found, with patch it should delete the version.

### DIFF
--- a/lib/hammer_cli_katello/content_view_purge.rb
+++ b/lib/hammer_cli_katello/content_view_purge.rb
@@ -84,7 +84,18 @@ module HammerCLIKatello
     private def old_unused_versions
       return @old_unused_versions if @old_unused_versions
 
-      all_versions = index(:content_view_versions, :content_view_id => options['option_id'])
+      all_versions = []
+      per_page = 100
+      page = 1
+
+      loop do
+        versions = index(:content_view_versions, :content_view_id => options['option_id'],
+                                                 :per_page => per_page, :page => page)
+        all_versions << versions
+        break if versions.count < per_page
+        page += 1
+      end
+      all_versions.flatten!
       all_versions.sort_by! { |v| [v['major'], v['minor']] }
       @old_unused_versions = all_versions.select do |v|
         v["environments"].empty? &&


lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -84,7 +84,8 @@ def execute
     private def old_unused_versions
       return @old_unused_versions if @old_unused_versions
 
-      all_versions = index(:content_view_versions, :content_view_id => options['option_id'])
+      all_versions = index(:content_view_versions, :content_view_id => options['option_id'],
+                                                   :per_page => 100)
```
ianballou, line 88: The original issue mentioned a number of content views even greater than 100.  Do you think it would make sense to loop over the content_view_versions `index` call to ensure even 1000 CVVs would get purged?  In the times when I've heard people using the purge command, they had an obscene amount of archived content view versions.

ianballou (commented):
The test steps worked fine for me, I had one comment about very large numbers of CVVs.

chris1984:
@ianballou bumped it from `100` to `1000` let me know if that looks good.

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -84,7 +84,8 @@ def execute
     private def old_unused_versions
       return @old_unused_versions if @old_unused_versions
 
-      all_versions = index(:content_view_versions, :content_view_id => options['option_id'])
+      all_versions = index(:content_view_versions, :content_view_id => options['option_id'],
```
ianballou, line 87: My worry about using a page size of 1000 is that it might lead to some very large API calls that could potentially time out or cause other issues.  Also, if the user for some reason has 2000 versions, they'll hit the bug again.

I posted an idea I had below. It seems to work okay. I loop through the content view versions, ensuring that all pages get accounted for.  That way, we get the benefit of a safely-sized API call too.

```suggestion
      all_versions = []
      per_page = 100
      page = 1

      loop do
        versions = index(:content_view_versions, :content_view_id => options['option_id'], :per_page => per_page, :page => page)
        all_versions << versions
        break if versions.count < per_page
        page += 1
      end
      all_versions.flatten!
```

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -84,7 +84,8 @@ def execute
...
+                                                   :per_page => 1000)
```
ianballou, line 88:
```suggestion
```

lib/hammer_cli_katello/content_view_purge.rb
chris1984: This looks better than my approach, will go with this ty.

ianballou (approved):
Did some extra testing, all looks good!

PR #884, opened 2023-03-28
Refs #36237 - add host info and update count totals on ak show

chris1984:
Needs https://github.com/Katello/katello/pull/10499

* Check out Hammer PR:
* Create an activation key
* Register a host with an activation key
* Run hammer activation-key info --id X and see if it shows the associated host
* Set a limit on your key and see if it shows how many more times it can be used

### DIFF
--- a/lib/hammer_cli_katello/activation_key.rb
+++ b/lib/hammer_cli_katello/activation_key.rb
@@ -55,7 +55,7 @@ module HammerCLIKatello
         field :name, _("Name")
         field :id, _("Id")
         field :description, _("Description"), Fields::Field, :hide_blank => true
-        field :format_limit, _("Host Limit")
+        field :format_consumed, _("Host Limit")
         field :auto_attach, _("Auto Attach")
         field :release_version, _("Release Version"), Fields::Field, :hide_blank => true
 
@@ -64,6 +64,11 @@ module HammerCLIKatello
         end
         from :content_view do
           field :name, _("Content View")
+        end
+
+        collection :hosts, _("Associated Hosts") do
+          field :id, _('Id')
+          field :name, _("Name")
         end
 
         collection :host_collections, _("Host Collections") do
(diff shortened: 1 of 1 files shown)


sjha4 (commented):
Tested with updated activation keys API:

Able to see hosts in the activation key details with hammer. ✔️ 
Able to see new formatted host limit in the format : Consumed of Host_limit ✔️ 

Screenshot from 2023-04-05 09-34-14

I am concerned about one thing here. If there are hundreds of hosts attached to an activation_key, the output may become unreadable. Should we provide an option to show hosts in the info here and not show them by default? Does that make sense?

chris1984:
> Tested with updated activation keys API:
> 
> Able to see hosts in the activation key details with hammer. ✔️ Able to see new formatted host limit in the format : Consumed of Host_limit ✔️
> 
> Screenshot from 2023-04-05 09-34-14
> 
> I am concerned about one thing here. If there are hundreds of hosts attached to an activation_key, the output may become unreadable. Should we provide an option to show hosts in the info here and not show them by default? Does that make sense?

Yeah that makes sense, let me see what I can do with it.

chris1984:
@sjha4 updated if you want to try again. pass in `--show-hosts` to see the hosts on the key

chris1984:
@sjha4 updated

Without hosts param:
```
[vagrant@hammer ~]$ hammer activation-key show --id 1
Name:                  Custom
Id:                    1
Host Limit:            1 of Unlimited
Auto Attach:           true
Lifecycle Environment: Library
Content View:          Default Organization View
Host Collections:

Content Overrides:

System Purpose:
    Service Level:
    Purpose Usage:
    Purpose Role:
    Purpose Addons:
```

With hosts param:
```
[vagrant@hammer ~]$ hammer activation-key show --show-hosts true --id 1
Name:                  Custom
Id:                    1
Host Limit:            1 of Unlimited
Auto Attach:           true
Lifecycle Environment: Library
Content View:          Default Organization View
Associated Hosts:
 1) Id:   1
    Name: devel.croberts.io
Host Collections:

Content Overrides:

System Purpose:
    Service Level:
    Purpose Usage:
    Purpose Role:
    Purpose Addons:
```

sjha4 (approved):
Works great 👍🏼

PR #891, opened 2023-05-16
Fixes #36355 - Added applied filters to hammer cli

qcjames53:
This PR attempts to fix Katello [#36355](https://projects.theforeman.org/issues/36355), a subtask of the Katello  [#36249](https://projects.theforeman.org/issues/36249) epic.

When querying info on a content view version, the returned CVV will now display a "Has Applied Filters" field; it will read either true, false, or blank. Blank represents a nil, occurring whenever a CVV was created before Katello #36251.

Sample input:
```
hammer content-view version info --id 10
```

Sample output:
```
Id:                     10
- - - - - [removed content] - - - - -
Has Applied Filters:    true
```

If the user adds the `--include-applied-filters true` option to their hammer command, the applied filter ids and names will be displayed, along with whether dependency solving is enabled.

Sample input:
```
hammer content-view version info --id 10 --include-applied-filters true
```

Sample output:
```
Id:                     10
- - - - - [removed content] - - - - -
Has Applied Filters:    true
Filters Applied:        
 1) :                   
     1) Name: Test Filer 1
        Id:   1
     2) Name: Test Filter 2
        Id:   2
     3) Name: Test Filter 3
        Id:   3
     4) Name: Test Filter 4
        Id:   4
     5) Name: Test Filter 5
        Id:   5
     6) Name: Test Filter 6
        Id:   6
    Dependency Solving: true
```


qcjames53:
This is for a command with `--include-applied-filters true`. The root "applied_filters" section would be absent if this option was omitted; the command output needs to handle this section not existing as well.

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -57,6 +55,17 @@ class InfoCommand < HammerCLIKatello::InfoCommand
           field :name, _("Name")
           field :label, _("Label")
         end
+
+        field :filters_applied, _("Has Applied Filters")
+        collection :applied_filters, _("Filters Applied") do
+          collection :applied_filters, _("") do
+            from :filter do
+              field :name, _("Name")
+              field :id, _("Id")
```
ofedoren, line 64: We always (at least I hope) show `id` first.
```suggestion
              field :id, _("Id")
              field :name, _("Name")
```

lib/hammer_cli_katello/content_view_version.rb
```diff
@@ -57,6 +55,17 @@ class InfoCommand < HammerCLIKatello::InfoCommand
           field :name, _("Name")
           field :label, _("Label")
         end
+
+        field :filters_applied, _("Has Applied Filters")
+        collection :applied_filters, _("Filters Applied") do
+          collection :applied_filters, _("") do
```
ofedoren, line 61: In case the structure will always be:
```ruby
{
  #...
  "filters_applied": true/false,
  "applied_filters": {
    "applied_filters": [
      {
        #...
      }
    ]
  }
}
```
With the first `applied_filters`  being redundant, then we can do:
```suggestion
        from :applied_filters,  do
          collection :applied_filters, _("Filters Applied"), hide_blank: true, hide_empty: true do
```

qcjames53: I was in a meeting yesterday with a couple people to address this, and the best we could come up with was the nested collection solution which slightly butchers the 'true' case but handles the 'false' and absent cases well for `--include-applied-filters`. Admittedly none of us were experts in this area.

qcjames53: Chris and I were able to fix this with a hide-blank on Dependency Solving. Thank you for the help with putting hide flags on the collection :)

chris1984 (approved):
ACK works great, thanks @qcjames53 for taking Oleh's advice as well.

ianballou:
@chris1984 @qcjames53 did you test this against filters that had rules in them? Since content view filters can be deleted after application, the user will likely need to look at the rules applied within the filters.

PR #897, opened 2023-06-28
Fixes #36546 - Update katello-agent dep warnings to 4.10

chris1984:


jeremylenz (approved):
Tested with katello-agent set up, works as expected.

ACK :+1:

PR #900, opened 2023-07-13
Run tests on Ruby 3 and remove LC_MESSAGES

ofedoren:
Not so important, but still useful.


chris1984:
Hey @ofedoren The tests are failing with ```ruby
WARNING: File locale/bn/LC_MESSAGES/hammer-cli-katello.mo does not exist, generate with 'make all-mo'!
No such file or directory @ rb_file_s_mtime - locale/bn/LC_MESSAGES/hammer-cli-katello.mo not found
wrong number of arguments (given 2, expected 1)```

Are you able to fix that up and we can merge this, I will make sure this gets in before branching.

ofedoren:
So outdated it hurts, closing since no longer needed.

PR #902, opened 2023-07-19
Fixes #36589 - Added support for calling ACS bulk actions through ham…

qcjames53:
…mer.

This was created to support a refresh-all feature request, in conjunction with [this Katello PR](https://github.com/Katello/katello/pull/10646).
[https://projects.theforeman.org/issues/36589](https://projects.theforeman.org/issues/36589)

### DIFF
--- a/lib/hammer_cli_katello/acs_bulk_actions.rb
+++ b/lib/hammer_cli_katello/acs_bulk_actions.rb
@@ -1,0 +1,38 @@
+module HammerCLIKatello
+  class AcsBulkActionsCommand < HammerCLIKatello::Command
+    desc 'Modify alternate content sources in bulk'
+    resource :alternate_content_sources_bulk_actions
+
+    class RefreshAllCommand < HammerCLIKatello::SingleResourceCommand
+      action :refresh_all_alternate_content_sources
+      command_name 'refresh-all'
+      desc _("Refresh all alternate content sources")
+      success_message _("Successfully refreshed all alternate content sources")
+      failure_message _("Could not refresh all alternate content sources")
+
+      build_options
+    end
+
+    class RefreshCommand < HammerCLIKatello::SingleResourceCommand
+      action :refresh_alternate_content_sources
+      command_name 'refresh'
+      desc _("Refresh alternate content sources")
+      success_message _("Successfully refreshed specified alternate content sources")
+      failure_message _("Could not refresh the specified alternate content sources")
+
+      build_options
+    end
+
+    class DestroyCommand < HammerCLIKatello::SingleResourceCommand
+      action :destroy_alternate_content_sources
+      command_name 'destroy'
+      desc _("Destroy alternate content sources")
+      success_message _("Sucessfully destroyed specified alternate content sources")
+      failure_message _("Could not destroy the specified alternate content sources")
+
+      build_options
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/acs_bulk_actions.rb
```diff
@@ -0,0 +1,38 @@
+module HammerCLIKatello
+  class AcsBulkActionsCommand < HammerCLIKatello::Command
+    desc 'Modify alternate content sources in bulk'
+    resource :alternate_content_sources_bulk_actions
+
+    class RefreshAllCommand < HammerCLIKatello::SingleResourceCommand
+      action :refresh_all_alternate_content_sources
```
ofedoren, line 7: Does this really work? I mean in the main Katello PR it's `/alternate_content_sources/bulk/refresh_all`, so here it should be `resource :bulk`, `action :refresh_all`...

qcjames53: It does. That's one of the things that was so tricky about the PR; hammer pulls the URL from the apidocs. If you click on the refresh all action in the APIdocs, at the top of the page it will show you the APIdoc location of the resource, `alternate_content_sources_bulk_actions/refresh_all_alternate_content_sources`. I think it pulls these from the function names in Katello.

qcjames53: At least that's the conclusion I came to after a bunch of testing the last few days ¯\_(ツ)_/¯

lib/hammer_cli_katello/acs_bulk_actions.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    class RefreshCommand < HammerCLIKatello::SingleResourceCommand
+      action :refresh_alternate_content_sources
+      command_name 'refresh'
+      desc _("Refresh alternate content sources")
+      success_message _("Successfully refreshed specified alternate content sources")
+      failure_message _("Could not refresh the specified alternate content sources")
+
+      build_options
+    end
```
sjha4, line 24: These SingleResourceCommand should be available in ACS command directly, no? Do we need these here?

qcjames53: In theory, yes. Since bulk ACS actions were a separate category in the API, I figured it would make sense to split this functionality off in a subcommand. If they were directly under the `alternate-content-source` command the refresh and destroy commands would need to be renamed something like `bulk-refresh` to not conflict with the existing ACS refresh and destroy commands.

lib/hammer_cli_katello/acs_bulk_actions.rb
qcjames53: Since the bulk refresh and destroy actions already existed, I don't think it makes sense to change API endpoints in Katello, at least right now.

sjha4 (approved):
New commands work great. Test based ack. 👍🏼

chris1984:
@ofedoren have all your concerns been addressed?

ofedoren:
Sorry, I've noticed this is waiting for me for some reason, so I don't want to block this. If it works, then I'm OK. I was just confused on why the resource name is so weird.

PR #906, opened 2023-08-29
[CP] Katello 4.10.0 CP for version 1.10.0

lfu:
hammer_cli_katello version has been bumped to 1.10.0.


ianballou:
Is hammer still complaining about the missing `.mo` files? If so, wonder if we could fix that as part of this CP.  @chris1984 might know how to fix it when he's back

ianballou:
I'm guessing the CP is fine if it was clean

lfu:
> Is hammer still complaining about the missing `.mo` files? 

Don't see that.

ianballou:
> > Is hammer still complaining about the missing `.mo` files?
> 
> Don't see that.

Gotcha, on master right now there are warnings about the `.mo` files being out of date when you run any command. If there are no warnings, then this should be good to merge

ianballou:
@parthaa do you wanna take a quick look since this is your PR that's CP'd in?

jeremylenz (approved):

PR #907, opened 2023-09-13
Fixes #36743 - Show full repository id list in hammer content-view list

sjha4:
`hammer content-view list` truncates repository list if there are a lot of repos in the cv.

The API returns the full list of repository ids, there's no pagination. Hammer however truncates the column if it's too long. Found the param max_width in hammer_cli to control how wide the column can be. Increasing to 300 but open to go higher if needed. 
Before PR:
Before_PR

After PR:
After_PR


chris1984 (approved):
Looks fine, based on your screenshot. I think 300 is good for now and if we get any complaints, we can bump it up or write a KCS article to let users know how to increase it.

PR #921, opened 2024-01-04
Fixes #37031 - hammer proxy content info fails when DB has no content counts

ianballou:
Test by upgrading from Katello 4.10 to 4.11 with a synced smart proxy. Run `hammer proxy content info...` and see the error "undefined method each for nil"

Related to https://github.com/Katello/katello/pull/10835/

### DIFF
--- a/lib/hammer_cli_katello/capsule.rb
+++ b/lib/hammer_cli_katello/capsule.rb
@@ -218,7 +218,7 @@ module HammerCLIKatello
           data["lifecycle_environments"].each do |lce|
             lce["content_views"].each do |cv|
               cv["repositories"].each do |repo|
-                if cv["up_to_date"]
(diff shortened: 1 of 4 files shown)


ianballou:
The test failure seems to be happening to any PR -- I think the GH ubuntu test machine is broken somehow. It's not reproducible locally on EL7.

chris1984:
Starting to test pr

chris1984 (approved):
Works great, thanks @ianballou I will figure out the test error, last time I saw this happen it was because of the seed order so half the time it would work, half the time it would fail. Here is what I did last time:

https://github.com/Katello/hammer-cli-katello/pull/910

One of the files was loading before another one and didn't have the proper require.

Feel free to merge it, since it's not related.

ianballou:
Thanks @chris1984 ! I'll take a look at what you did before and see if I can't fix it with this PR.

PR #922, opened 2024-01-15
Fixes #37053 - Add CV/LCE back host list params

chris1984:
What are the changes introduced in this pull request?
* With the new `content_view_environments` collection we were no longer pulling the data correctly; this has been updated. 

Considerations taken when implementing this change?

* This might not work with multiple cv/lce's since `hammer list` does not support using `collection` only the `info` command does. If this does not work with multiple cve's then when we get to that point, we will have to raise a redmine to `hammer-cli core` to add support to use collection with the list command. 

What are the testing steps for this pull request?
* Check out PR
* Register a host to your Katello devel/production box
* run `hammer host list` and see if the host Content view and Lifecycle environment fields are back

Before PR:
```bash
---|-------------------|------------------|------------|----------------|-------------------|--------------
ID | NAME              | OPERATING SYSTEM | HOST GROUP | IP             | MAC               | GLOBAL STATUS
---|-------------------|------------------|------------|----------------|-------------------|--------------
12 | devel.croberts.io | CentOS_Stream 8  |            | 192.168.41.128 | 00:0c:29:82:76:c6 | Warning
---|-------------------|------------------|------------|----------------|-------------------|--------------
```

With PR:
```bash
---|-------------------|------------------|------------|----------------|-------------------|---------------|---------------------------|----------------------
ID | NAME              | OPERATING SYSTEM | HOST GROUP | IP             | MAC               | GLOBAL STATUS | CONTENT VIEW              | LIFECYCLE ENVIRONMENT
---|-------------------|------------------|------------|----------------|-------------------|---------------|---------------------------|----------------------
12 | devel.croberts.io | CentOS_Stream 8  |            | 192.168.41.128 | 00:0c:29:82:76:c6 | Warning       | Default Organization View | Library
---|-------------------|------------------|------------|----------------|-------------------|---------------|---------------------------|----------------------
```

### DIFF
--- a/lib/hammer_cli_katello/host_extensions.rb
+++ b/lib/hammer_cli_katello/host_extensions.rb
@@ -26,8 +26,12 @@ module HammerCLIKatello
     ::HammerCLIForeman::Host::ListCommand.instance_eval do
       output do
         from :content_facet_attributes do
-          field :content_view_name, _('Content View')
-          field :lifecycle_environment_name, _('Lifecycle Environment')
+          from :content_view do
+            field :name, _('Content View'), Fields::List
+          end
+          from :lifecycle_environment do
+            field :name, _('Lifecycle environment'), Fields::List
+          end
           from :errata_counts do
             field :security, _("Security"), nil, :sets => ['ALL']
             field :bugfix, _("Bugfix"), nil, :sets => ['ALL']


qcjames53 (approved):
I tested changes with a variety of hosts with different properties, this seems to work as advertised across the board. Ack'ing.

I didn't know lists do not support collections, but that makes sense. Since other elements in hammer do not show fields when they are one-to-many, we could end up omitting this field in the future. Another idea: perhaps we could display a "(multiple)" string if there is more than one CV or LE on a host?

chris1984:
> I tested changes with a variety of hosts with different properties, this seems to work as advertised across the board. Ack'ing.
> 
> I didn't know lists do not support collections, but that makes sense. Since other elements in hammer do not show fields when they are one-to-many, we could end up omitting this field in the future. Another idea: perhaps we could display a "(multiple)" string if there is more than one CV or LE on a host?

That is a better approach we can take, so we don't have to rely on hammer-core and get blocked if they are busy. Will do that when we get back to doing the multi CV stuff. Thanks for the review!!

PR #923, opened 2024-01-22
Use foreman GH action workflow

chris1984:


chris1984:
@evgeni do you know why this is failing?

.github/workflows/unit-tests.yml
```diff
@@ -1,27 +1,18 @@
...
-      with:
-        ruby-version: ${{ matrix.ruby }}
-        bundler-cache: true
-    - name: Run tests
-      run: bundle exec rake
+  test:
+    name: Tests
+    uses: theforeman/actions/.github/workflows/test-gem.yml@v0
+    with:
+      command: bundle exec rake test
```
ekohl, line 18: I'd ensure a newline here.

.github/workflows/unit-tests.yml
```diff
@@ -1,27 +1,18 @@
...
+  rubocop:
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
 
-    steps:
-    - uses: actions/checkout@v2
-    - name: Add hammer-cli-foreman to local gem file
-      run: echo "gem 'hammer_cli_foreman', :git => 'https://github.com/theforeman/hammer-cli-foreman.git'" > Gemfile.local
-    - name: Add hammer-cli to local gem file
-      run: echo "gem 'hammer_cli', :git => 'https://github.com/theforeman/hammer-cli.git'" >> Gemfile.local
```
ekohl, line 20: These are now only pulled in via released gems, no longer the git version. If you care about this, I'd recommend adding it to your Gemfile. If you want, you can limit it to `ENV['GITHUB_ACTIONS'] == 'true'`.

ekohl (changes requested):
It's failing on a few things:

* Ruby 3.2 fails because Rake 10 doesn't work with it:
https://github.com/Katello/hammer-cli-katello/blob/9a6ae05cd72e8dd09004204c00571d449abe4c63/Gemfile#L9
* Ruby 3.1 - 2.6 fail on `gem build --verbose --strict *.gemspec`. Some of those can be resolved by merging the `gemspec` with `Gemfile` dependencies.
* Ruby 2.5 - 2.4 fail because there is no `--strict` parameter to `gem build` in those versions. I'd add a minimum Ruby version of 2.7 in the gemspec to avoid testing on those old versions.

hammer_cli_katello.gemspec
```diff
@@ -74,7 +75,8 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency 'simplecov'
   gem.add_development_dependency 'mocha'
   gem.add_development_dependency 'ci_reporter'
-
-  gem.add_development_dependency "rubocop", "0.42"
-  gem.add_development_dependency "rubocop-checkstyle_formatter"
+  gem.add_development_dependency 'rubocop', '0.42'
```
ekohl, line 78: This is a seriously old version. It wouldn't surprise me if this wouldn't work on Ruby 3.

hammer_cli_katello.gemspec
```diff
@@ -74,7 +75,8 @@ Gem::Specification.new do |gem|
...
+  gem.add_development_dependency 'rubocop-checkstyle_formatter'
+  # for generating i18n files, gettext > 3.0 dropped ruby 1.8 support
```
ekohl, line 80: I think the Ruby 1.8 part isn't that relevant anymore.

hammer_cli_katello.gemspec
```diff
@@ -74,7 +75,8 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency 'simplecov'
   gem.add_development_dependency 'mocha'
   gem.add_development_dependency 'ci_reporter'
-
-  gem.add_development_dependency "rubocop", "0.42"
-  gem.add_development_dependency "rubocop-checkstyle_formatter"
+  gem.add_development_dependency 'rubocop', '0.42'
+  gem.add_development_dependency 'rubocop-checkstyle_formatter'
```
ekohl, line 79: This was only used for testing in Jenkins, but even that doesn't need it anymore.
```suggestion
```

Gemfile
```diff
@@ -2,21 +2,6 @@ source "https://rubygems.org"
 
 gemspec
 
-# for generating i18n files, gettext > 3.0 dropped ruby 1.8 support
-gem 'gettext', '>= 3.1.3', '< 4.0.0'
-
-group :test do
```
ekohl, line 8: We install RuboCop testing without development dependencies, but with test dependencies. So just moving everything to a development dependency broke the RuboCop.

chris1984: Ah ok that makes, sense, will fix it up. Thanks for looking again

chris1984: Ah so these all would need to be done as:
`gem.add_dependency` then since I was looking at the docs for `gemspec` and you can't do grouping like you could in the `Gemfile`

If that is correct, I will go ahead and get it changed up

evgeni: Yeah, in a gemspec you can only do `add_dependency` (which is runtime) and `add_development_dependency` which is devel and maps to the `development` bundler group from the gemfile

hammer_cli_katello.gemspec
```diff
@@ -63,18 +63,19 @@ Gem::Specification.new do |gem|
   gem.name = 'hammer_cli_katello'
   gem.require_paths = ['lib']
   gem.version = HammerCLIKatello.version
+  gem.required_ruby_version = '>= 2.7'
 
   gem.add_dependency 'hammer_cli_foreman'
   gem.add_dependency 'hammer_cli_foreman_tasks'
+  gem.add_dependency 'gettext', '>= 3.1.3', '< 4.0.0'
```
evgeni, line 70: this is a devel dependency I think?

evgeni:
https://github.com/Katello/hammer-cli-katello/pull/924 is turning green now, see the individual commits in that PR for an explanation why/what has happened there.

Gemfile
evgeni: fwiw, @ekohl is wrong, we do not skip `development` group when installing rubocop: https://github.com/theforeman/actions/blob/v0/.github/workflows/rubocop.yml

chris1984:
> #924 is turning green now, see the individual commits in that PR for an explanation why/what has happened there.

Thanks for putting them into single commits, I see what I was doing wrong and what you meant now about my question about Ewoud's comment and then yours. This helps me do hammer azure/virt-who/rh_cloud so appreciate you doing this one as an example. Closing my pr out in favor of yours.

PR #924, opened 2024-01-25
Use foreman GH action workflow

evgeni:


chris1984 (approved):
Thanks @evgeni will model the other hammer ones off this one. Learned a few things as well :)

evgeni:
One thing you might want to do first (here) is to see whether all the dev deps are actually needed and/or exist in newer versions.

hammer_cli_katello.gemspec
```diff
@@ -63,18 +63,17 @@ Gem::Specification.new do |gem|
...
-  gem.add_development_dependency 'minitest-spec-context'
-  gem.add_development_dependency 'simplecov'
-  gem.add_development_dependency 'mocha'
-  gem.add_development_dependency 'ci_reporter'
-
-  gem.add_development_dependency "rubocop", "0.42"
-  gem.add_development_dependency "rubocop-checkstyle_formatter"
+  gem.add_development_dependency 'minitest-spec-context', '~> 0.0.5'
+  gem.add_development_dependency 'mocha', '~> 2.0'
+  gem.add_development_dependency 'rake', '~> 10.0'
```
ekohl, line 76: I think this broke on Jenkins:
```
21:27:43  /usr/local/rvm/gems/ruby-2.7.4@hammer-cli-katello-master-source-release-193-1-ruby-2-7/gems/bundler-2.4.22/lib/bundler/rubygems_integration.rb:308:in `block in replace_bin_path': can't find executable rake for gem rake. rake is not currently included in the bundle, perhaps you meant to add it to your Gemfile? (Gem::Exception)
```

Couldn't you have bumped to '~> 13.0`?

evgeni: interesting. and no, see https://github.com/Katello/hammer-cli-katello/pull/924/commits/f66bd758ca963872c5f2a02eaed7fa9169a88290

evgeni: 11+ does something that our tests start failing

ekohl: Regardless, Jenkins still doesn't install the development dependencies so a test group is still useful. Or just list rake unconditionally in `Gemfile`.

ekohl: > 11+ does something that our tests start failing

Yes, it starts displaying warnings.

I couldn't find the source of the incorrect interpolation warning, which certainly is a bug. I could trace it to https://github.com/theforeman/hammer-cli/blob/f38afde3e44667908ed39078c48fb26a2244aa8b/lib/hammer_cli/utils.rb#L15 but didn't find a stack trace of where it's called from. It appears (at least on Ruby 3.2) to end up calling `string % nil` which emits a warning.

I at least opened https://github.com/Katello/hammer-cli-katello/pull/925 for some other bugs I noticed.

ofedoren: Just to mention this here as well: I'd really try to use either unpinned `rake` or at least as Ewoud suggested `~> 13.0`. Hammer core does this and it seems to be working for quite some time?

ekohl: This gem raises many warnings, mostly about redefining environment, but also at least one string formatting warning.

chris1984: I will bump it to `~> 13.0` as Ewoud suggested and see what else we can drop or bump from here

ekohl: > Regardless, Jenkins still doesn't install the development dependencies so a test group is still useful. Or just list rake unconditionally in `Gemfile`.

This is still true and nightlies for this package have been broken since. Is there still a desire for nightly packages?

PR #925, opened 2024-01-26
Fix broken string interpolation with translations

ekohl:
Previously it was interpolating a string and then translating it. That never works. Instead, you must first translate it and then interpolate.


chris1984 (approved):
Thanks @ekohl

PR #933, opened 2024-03-29
Fixes #37311 - Add capsule content repair command

sjha4:
Test with: https://github.com/Katello/katello/pull/10932

Screenshot from 2024-03-29 12-04-02


chris1984 (approved):
Works great :) Tested it on a stream box and works well with the Katello PR

PR #934, opened 2024-04-25
Add packit support

ianballou:
I used hammer-cli-foreman's packit file as an example.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,3 +7,13 @@ gemspec
   local_gemfile = File.join(File.dirname(__FILE__), file_name)
   self.instance_eval(Bundler.read_file(local_gemfile)) if File.exist?(local_gemfile)
 end
+
+group :test do
+  gem 'gettext', '>= 3.1.3', '< 4.0.0'
+  gem 'minitest', '4.7.4'
+  gem 'minitest-spec-context', '~> 0.0.5'
+  gem 'mocha', '~> 2.0'
+  gem 'rake', '~> 10.0'
+  gem 'rubocop', '0.42'
+  gem 'thor', '~> 1.0'
+end
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,47 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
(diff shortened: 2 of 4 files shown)


chris1984 (approved):
ACK

ianballou:
https://github.com/theforeman/foreman-packaging/pull/10731 will fix the builds

gemfile.d/test.rb
```diff
@@ -0,0 +1,8 @@
+group :test do
+  gem 'ci_reporter', '>= 1.6.3', "< 2.0.0"
```
ekohl, line 2: This one was for Jenkins. I don't think it supports GHA

Gemfile
```diff
@@ -7,3 +7,7 @@ gemspec
   local_gemfile = File.join(File.dirname(__FILE__), file_name)
   self.instance_eval(Bundler.read_file(local_gemfile)) if File.exist?(local_gemfile)
 end
+
+Dir[File.join(__dir__, 'gemfile.d', '*.rb')].each do |bundle|
```
ekohl, line 11: There's no need for this here. You can just define it in the Gemfile itself. Gemfile.d is for Foreman plugins.

While we don't use it today, dependabot doesn't support `eval_gemfile` so I'd stick to a plain Gemfile.

Though the instance_eval above already breaks dependabot.

ekohl (changes requested):
I'd also drop the test deps from the gemspec to avoid duplication

ianballou:
This PR should fix our nightly builds, which folks are waiting to test.

gemfile.d/test.rb
ekohl: If you're dropping the use of it, you can also drop the dependency
```suggestion
```

ianballou:
@ekohl thanks for catching the gemfile folder, it's deleted now.

ianballou:
Did a quick test, my PR doesn't seem to break anything in h-c-k at least.

chris1984 (approved):

ianballou:
Merging since it _should_ help with nightly, but we can make further improvements as needed.

PR #936, opened 2024-05-08
Add ci_reporter_minitest to Gemfile and match h-c-f more closely

ianballou:
Should fix the issue in nightly: `[2024-05-07T19:54:51.653Z] Don't know how to build task 'ci:setup:minitest' (see --tasks)`

Also made some updates to better match https://github.com/theforeman/hammer-cli-foreman/blob/master/Gemfile

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,9 +8,11 @@ gemspec
   self.instance_eval(Bundler.read_file(local_gemfile)) if File.exist?(local_gemfile)
 end
 
+gem 'gettext', '>= 3.1.3', '< 4.0.0'
+
 group :test do
-  gem 'gettext', '>= 3.1.3', '< 4.0.0'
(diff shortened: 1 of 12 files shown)


chris1984 (changes requested):
A rubocop failure and are those test failures related?

ianballou:
I upgraded mini test, this might be the fallout from that. I'll take a look into what needs fixing.

ianballou:
I needed to remove the use of `let(:name)` due to:

```
/home/vagrant/hammer-cli-katello/.vendor/ruby/3.0.0/gems/minitest-5.22.3/lib/minitest/spec.rb:253:in `let': let 'name' cannot override a method in Minitest::Spec. Please use another name. (ArgumentError)
```

chris1984 (approved):
Looks good

PR #943, opened 2024-05-28
Use theforeman-rubocop gem

archanaserver:
Choose to inherit from lenient.yml, because wanted to inherit rules from style_lenient.yml as well, i followed this:
https://github.com/theforeman/theforeman-rubocop?tab=readme-ov-file#not-intrusive-style---lenient

Also drop some cops from the .rubocop.yml config, because that was present in inherit-gem and added theforeman-rubocop gem.

This is part of Rubocop standerdization, link for the reference: https://community.theforeman.org/t/standardizing-rubocop-with-theforeman-rubocop/37239


chris1984 (changes requested):
Looks like some files need to be updated with the new gem, can you submit them in a separate commit in this pr?

chris1984 (changes requested):
Can this one be rebased with your fixes from your other pr then this should be green

chris1984 (changes requested):
@archanaserver acked your other pr, can you rebase this one to resolve the conflicts

archanaserver:
Thanks @chris1984 and this one is green now!

chris1984 (approved):

hammer_cli_katello.gemspec
```diff
@@ -65,6 +65,8 @@ Gem::Specification.new do |gem|
   gem.version = HammerCLIKatello.version
   gem.required_ruby_version = '>= 2.7', '< 3.2'
 
+  gem.add_dependency 'theforeman-rubocop', '~> 0.1.0'
```
ekohl, line 68: This should have been a development dependency, not a runtime dependency

evgeni: https://github.com/Katello/hammer-cli-katello/pull/950

PR #945, opened 2024-08-09
Fixes #37721 - Hammer erratum list output is wrong when using content-view filter

sjha4:
Add some validations around --content-view filter for errata lists.

1. If you're using cv-name, hammer needs to know org id to fetch the cv-id
2. Errata belongs to cv versions and not cv directly so if you're using --content-view(-id) filter, you also need to provide --environment params to fetch the exact cvv.
3. The --environment param was not working. Only the environment-id param was. This should fix that.

To test this create a CV with repo with some errata and publish it to a version promoted to any environment.

Then run:
hammer erratum list --content-view='RHEL7' --cve="CVE-2023-45288" --lifecycle-environment-id=1
 with different params passed to m,ake sure it all looks ok.

### DIFF
--- a/lib/hammer_cli_katello/erratum.rb
+++ b/lib/hammer_cli_katello/erratum.rb
@@ -22,16 +22,27 @@ module HammerCLIKatello
       validate_options :before, 'IdResolution' do
         organization_options = [:option_organization_id, :option_organization_name, \
                                 :option_organization_label]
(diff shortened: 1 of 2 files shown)


ianballou (approved):
Working well! I tested the scenarios in the redmine and then some more on top of that. 

Also double checked that the content-view-filter + available-for filters still work, just to be safe.

PR #947, opened 2024-08-20
Bump version to 1.15.0-main

chris1984:


lib/hammer_cli_katello/version.rb
```diff
@@ -1,5 +1,5 @@
 module HammerCLIKatello
   def self.version
-    @version ||= Gem::Version.new('1.14.0-master')
+    @version ||= Gem::Version.new('1.15.0-master')
```
evgeni, line 3:
```suggestion
    @version ||= Gem::Version.new('1.15.0-main')
```

your branch is called `main`, so should your gem, or you're gona have a bad time :(

chris1984: Damn, I'll revert and fix them. Now I agree with Ewoud we should have just left the branch name

evgeni: Nah, we just now know what all needs to be touched when we rename katello;)

PR #950, opened 2024-08-21
make theforeman-rubocop a development dependency

evgeni:


ofedoren (approved):

ekohl (changes requested):
It also lists rubocop explicitly in the test group:
https://github.com/Katello/hammer-cli-katello/blob/c53c59a60de37c939c0b7b386aafabf40149cc19/Gemfile#L19

Feels like this should be unified.

ehelms (approved):

PR #954, opened 2024-09-02
Declare compatible with Ruby < 4

ekohl:
This helps developers on current Fedora where either Ruby 3.2 (f39) or Ruby 3.3 (f40) is used.


chris1984 (approved):
Thanks @ekohl

PR #956, opened 2024-09-27
Add release workflow

chris1984:


.github/workflows/release.yml
```diff
@@ -0,0 +1,16 @@
...
+  push:
+    tags:
+      - '*'
+
+jobs:
+  release:
+    name: Release gem
+    uses: theforeman/actions/.github/workflows/release-gem.yml@v0
+    with:
+      allowed_owner: theforeman
```
evgeni, line 14:
```suggestion
      allowed_owner: Katello
```

Unless you wanted to migrate that repo over to @theforeman :)

chris1984: Updated and changed the secret to the key you sent me in a PM

evgeni (approved):

PR #958, opened 2024-10-01
Fixes #37865 - Add Multi environments to activation key info

chris1984:
* Adds support for multi environment activation keys.
* Enabled a test we had skipped since Katello 4.5 that has been fixed for a long time
* Needs this Katello PR to work - https://github.com/Katello/katello/pull/11166

With PR with an ak that has multiple envs:
```bash
[vagrant@devel hammer-cli-katello]$ hammer activation-key info --id 36
Name:                           foo-env
Id:                             36
Host Limit:                     0 of Unlimited
Multi Content View Environment: yes
Organization:
 1) Id:   1
    Name: Default Organization
Content view environments:
 1) Content view:
        Id:                      2
        Name:                    Mixed
        Version:                 1.0
        Content view version Id: 2
        Composite:               no
    Lifecycle environment:
        Id:   1
        Name: Library
    Candlepin Name:        Library/Mixed
 2) Content view:
        Id:                      3
        Name:                    Animals
        Version:                 1.0
        Content view version Id: 3
        Composite:               no
    Lifecycle environment:
        Id:   2
        Name: Dev
    Candlepin Name:        Dev/Animals
```
AK with single env:
```bash
[vagrant@devel hammer-cli-katello]$ hammer activation-key info --id 35
Name:                           foo
Id:                             35
Host Limit:                     0 of Unlimited
Multi Content View Environment: no
Organization:
 1) Id:   1
    Name: Default Organization
Content view environments:
 1) Content view:
        Id:                      2
        Name:                    Mixed
        Version:                 1.0
        Content view version Id: 2
        Composite:               no
    Lifecycle environment:
        Id:   1
        Name: Library
    Candlepin Name:        Library/Mixed
```


jeremylenz:
What do you think about adding another field like this?

```
Content View Environments: Library/Mixed, Dev/Animals
```

This would provide an easy way to copy/paste the current CVEs somewhere so you could edit as needed. It would also need another rabl node added in Katello.

Also it struck me as a bit weird that we don't provide each CVE's id anywhere. do you think we should add that as well?

chris1984:
> What do you think about adding another field like this?
> 
> ```
> Content View Environments: Library/Mixed,Dev/Animals
> ```
> 
> This would provide an easy way to copy/paste the current CVEs somewhere so you could edit as needed. It would also need another rabl node added in Katello.
> 
> Also it struck me as a bit weird that we don't provide each CVE's id anywhere. do you think we should add that as well? This way it could help with `hammer activation-key update --content-view-environment-ids xxx`

+1 to those ideas, when you say another rabl node would it go in https://github.com/Katello/katello/blob/bc39167a85e8aaa3ab4285bd459e672c7bcfc7a5/app/views/katello/api/v2/activation_keys/base.json.rabl or somewhere else? It sounds like this is related to that other card

jeremylenz:
> would it go in https://github.com/Katello/katello/blob/bc39167a85e8aaa3ab4285bd459e672c7bcfc7a5/app/views/katello/api/v2/activation_keys/base.json.rabl or somewhere else?

Yes that's the file where it would go. 

I'm realizing we already have a node called `:content_view_environments` which is nested, so it'd have to be called something else. Maybe `:content_view_environment_labels`

parthaa:
> +1 to those ideas, when you say another rabl node would it go in https://github.com/Katello/katello/blob/bc39167a85e8aaa3ab4285bd459e672c7bcfc7a5/app/views/katello/api/v2/activation_keys/base.json.rabl or somewhere else? It sounds like this is related to that other card

Doesnt need to be in RABL why not just in the info command?

chris1984:
@jeremylenz updated this pr and the Katello one, when you get time to test it

jeremylenz:
When I do hammer activation-key list I still see
```
---|--------|----------------|-----------------------|-------------
ID | NAME   | HOST LIMIT     | LIFECYCLE ENVIRONMENT | CONTENT VIEW
---|--------|----------------|-----------------------|-------------
1  | ak1    | 1 of Unlimited | Library               | cv1         
4  | ak3    | 0 of Unlimited | Library               | cv1         
3  | akZero | 0 of Unlimited |                       |             
---|--------|----------------|-----------------------|-------------
```

I think we should replace the lifecycle environment and content view columns with 'Content view environments' and use the new field you made.

lib/hammer_cli_katello/activation_key.rb
```diff
@@ -56,41 +56,63 @@ def request_params
...
+              field :content_view_environment_id, _("Content view environment id"), Fields::Field
+            end
+          end
+          from :lifecycle_environment do
+            label _("Lifecycle environment") do
+              field :id, _("Id")
+              field :name, _("Name")
+            end
+          end
+          field :label, _("Environment Label")
```
jeremylenz, line 85: Saying 'Environment Label' here will imply that it's about lifecycle environment.

```suggestion
          field :label, _("Label")
```

lib/hammer_cli_katello/activation_key.rb
```diff
@@ -56,41 +56,63 @@ def request_params
...
-        from :environment do
-          field :name, _("Lifecycle Environment")
+        collection :organization, _("Organization") do
+          field :id, _("Id"), Fields::Field, :hide_blank => true
+          field :name, _("Name"), Fields::Field, :hide_blank => true
         end
-        from :content_view do
-          field :name, _("Content View")
+
+        collection :content_view_environments, _('Content view environments') do
```
jeremylenz, line 68: I do like sentence case more than title case, but we need to be consistent throughout Hammer output. 

```suggestion
        collection :content_view_environments, _('Content View Environments') do
```

chris1984:
@jeremylenz updated! Fixed the list command too
```bash
---|-----------|----------------|----------------------------|-------------------------------
ID | NAME      | HOST LIMIT     | CONTENT VIEW ENVIRONMENTS  | MULTI CONTENT VIEW ENVIRONMENT
---|-----------|----------------|----------------------------|-------------------------------
3  | multi-key | 0 of Unlimited | Library/RHEL-8,Dev/Animals | yes
2  | RHEL-9    | 1 of Unlimited | Library/RHEL-9             | no
1  | Rocky-8   | 1 of Unlimited | Library/Rocky-8            | no
---|-----------|----------------|----------------------------|-------------------------------
```

jeremylenz (approved):
LGTM 👍

jeremylenz (approved):

PR #968, opened 2024-11-07
Fixes #37993 - Update cvpurge count to a better description

chris1984:
* The `--count` parameter was causing confusion amongst users which lead them to believe that is how many versions to purge instead of keep.
* Renamed the param to `--versions-to-keep` and updated the options in the code and in the tests.
* The code looks like this which determines how many versions to delete `versions_to_purge = old_unused_versions.slice(0, old_unused_versions.size - option_versions_to_keep)`

### To test
* Create a content view and publish 4 versions
* If you don't pase in `versions-to-keep` the default is 3 it will remove 1 version
* Create a few more published versions of that content view
* Pass in how many you want to keep and issue the command to make sure it deletes the versions that are unused minus how many you want to keep.

### DIFF
--- a/lib/hammer_cli_katello/content_view_purge.rb
+++ b/lib/hammer_cli_katello/content_view_purge.rb
@@ -7,9 +7,11 @@ module HammerCLIKatello
     command_name "purge"
     desc "Delete old versions of a content view"
 
-    option "--id", "ID", _("Content View numeric identifier")
-    option "--name", "NAME", _("Content View name")
-    option "--count", "COUNT", _("count of unused versions to keep"),
+    option "--id", "Id", _("Content View numeric identifier")
+    option "--name", "Name", _("Content View name")
+    option "--count", "Count", _("(deprecated) Number of versions to keep"),
+           :deprecated => true, format: HammerCLI::Options::Normalizers::Number.new
+    option "--versions-to-keep", "Versions to keep", _("Number of unused versions to keep"),
            default: 3, format: HammerCLI::Options::Normalizers::Number.new
 
     validate_options :before, 'IdResolution' do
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -48,18 +48,19 @@ def option_sources
       sources
     end
 
+    # rubocop:disable Layout/LineLength
```
qcjames53, line 51: If you wouldn't mind, please remove this linelength limit and split the long lines 54 and 63 into smaller lines.

test/functional/content_view/purge_test.rb
```diff
@@ -52,16 +52,18 @@ module HammerCLIKatello
     it 'fails gracefully if count <= 0' do
       api_expects_no_call
 
-      r = run_cmd(%w(content-view purge --id 2 --count -1))
-      assert(r.err.include?('Invalid value for --count option'), 'Incorrect error message')
+      r = run_cmd(%w(content-view purge --id 2 --versions-to-keep -1))
+      # rubocop:disable Layout/LineLength
```
qcjames53, line 56: Same here for 57, please:
```
      assert(r.err.include?(
        'Invalid value for --versions-to-keep option'),
        'Incorrect error message'
      )
```

qcjames53 (changes requested):
Hi Chris! I was able to verify that the hammer field works properly with a variety of inputs. I could see it still being confusing for users that the number of CVVs remaining is one (or more) more than the "versions to keep" flag they passed in, but I do think the help text does a good job of showing that it's the unused CVVs. Since this isn't changing any of the functionality of this command, I'm totally okay with how purge functions.

The source looks good to me but I'd love if you could split the long source lines instead of disabling the limit.

lib/hammer_cli_katello/content_view_purge.rb
```diff
@@ -7,9 +7,9 @@ class ContentViewPurgeCommand < HammerCLIKatello::Command
     command_name "purge"
     desc "Delete old versions of a content view"
 
-    option "--id", "ID", _("Content View numeric identifier")
-    option "--name", "NAME", _("Content View name")
-    option "--count", "COUNT", _("count of unused versions to keep"),
```
parthaa, line 12: check on deprecation

qcjames53:
Wanted to get this down in writing before I forget:

We discussed changing the linelength config for rubocop hammer to match rubocop katello because they are so different. Katello is set up for length 200 at the moment. I'm totally okay with these changes and that would let the longer lines here through without modification.

chris1984:
@qcjames53 how does this look:
#### Help
```bash
[vagrant@ip-10-0-168-55 hammer-cli-katello]$ hammer content-view purge -h
Usage:
    hammer content-view purge [OPTIONS]

Options:
 --async                          Do not wait for the task
 --count NUMBER                   (deprecated) Number of versions to keep
                                  Default: 3
 --id VALUE                       Content View numeric identifier
 --name VALUE                     Content View name
 --organization[-id|-label] VALUE Organization name/label/id to search by
 --versions-to-keep NUMBER        Number of unused versions to keep
                                  Default: 3
 -h, --help                       Print help
```

#### Command showing the deprecation warning:
```bash
[vagrant@ip-10-0-168-55 hammer-cli-katello]$ hammer content-view purge --organization-id 1 --id 5 --count 3
The --count option is deprecated and will be removed in the next release.
[.................................................................................................................................................................................................................................................................................................................................................................................] [100%]
Version '1.0' of content view 'Test-Purge' deleted.
```

qcjames53 (approved):
The new help text looks great to me; both options have appropriate descriptions. I was able to trigger the deprecation warning when running `--count`. Functionality for both purge options seems to work. The unit tests seem to work well. ACKing :)

PR #970, opened 2024-11-14
Fixes #38014 - Add safe navigation for empty capsule content counts

chris1984:
* There is an action in Katello that can cause capsule content counts to be an empty hash, we don't safely navigate in Ruby to handle this and hammer blows up with a traceback causing a bad user experience. @sjha4 can probably add more context on how it happens.

* Added some unit tests to account for the empty hash and to make sure we handle it correctly

* Spin up a Satellite/Capsule or if you have a katello devel box add a smartproxy to it
* Have a Satellite with external Capsule, turn automatic content count calculation off.
* Create an LCE, assign it to the Capsule.
* Create and sync some repo (fake_yum1)
* Create a CV, add the repo to it, publish and promote to the LCE -> Capsule gets autosynced.
* On the Satellite/Devel box run the following:
* `foo = SmartProxy.find(X)`
* `foo.content_counts = {}`
* `foo.save`
* On Satellite try `hammer capsule content info --id X
* Apply PR and see if it shows the Capsule content info without a traceback

### DIFF
--- a/lib/hammer_cli_katello/capsule.rb
+++ b/lib/hammer_cli_katello/capsule.rb
@@ -248,14 +248,16 @@ module HammerCLIKatello
 
         private
 
+        # rubocop:disable Metrics/CyclomaticComplexity
+        # rubocop:disable Metrics/PerceivedComplexity
         def extend_data(data)
-          data["lifecycle_environments"].each do |lce|
-            lce["content_views"].each do |cv|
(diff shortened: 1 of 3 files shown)


sjha4 (approved):
Ack..Tested with content_counts populated, nil and {}. Output is formatted correctly for all cases. 💚

PR #973, opened 2024-11-26
Fixes #37151 - Add --force flag to content overrides

chris1984:
* We only support enabled as a value on content overrides, adding a flag that makes the user add --force for anything other than enabled. 
* Increased the line length to 150 from discussions with Quinn on my last PR since Katello is at 200.
* Added some unit tests around using the new `--force` option and validating we error out with invalid usage if the flag is not present. 

#### Considerations taken when implementing this change?
* Found some left over stuff from syspurpose addons I had to fix for rubocop, will remove all of that in a separate pr since it is now removed from katello.

* Try changing a content over ride on a host subscription `hammer host subscription content-override --override-name food --value disable --content-label test --host-id 3` 
* Apply PR 
* Try the same command see that it prompts you for the `--force` flag 
* Try with `--force` and it should go through
* Same with `hammer activation-key content-override --override-name food --value disable --content-label test --id 1` I am not sure how content labels work for activation keys since I get `Content label(s) "test" were not found in the Organization "Default Organization"` but the pr still works to require the user to pass in `--force`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -22,7 +22,7 @@ Encoding:
   Enabled: false
 
 LineLength:
-  Max: 100
+  Max: 150
 
 FormatString:
   Enabled: false # we use % for i18n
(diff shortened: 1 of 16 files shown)


chris1984:
Can't reopen pr thanks GH for closing it 😠 https://github.com/Katello/hammer-cli-katello/pull/977 is the new one

PR #977, opened 2024-12-03
Fixes #37151 - Add --force flag to content overrides

chris1984:
We only support enabled as a value on content overrides, adding a flag that makes the user add --force for anything other than enabled. Increased the line length to 150 from discussions with Quinn on my last PR since Katello is at 200.

@jeremylenz github was down earlier and was stuck pushing my changes, after it finished 30 mins later it closed my pr and I can't open it again saying the branch was force pushed to. So here is the updated pr with the changes you requested in https://github.com/Katello/hammer-cli-katello/pull/973

### DIFF
--- a/test/functional/host/subscription/content_override_test.rb
+++ b/test/functional/host/subscription/content_override_test.rb
@@ -9,9 +9,9 @@ describe 'host subscription content-override' do
     @cmd = %w(host subscription content-override)
   end
 
-  it "attaches a content label" do
+  it "attaches a content override" do
     label = "foo"
-    value = 'default'
+    value = 'enabled'
     id = '20'
     params = ["--host-id=#{id}", "--content-label=#{label}", "--value=#{value}"]
     ex = api_expects(:host_subscriptions, :content_override, "content override") do |par|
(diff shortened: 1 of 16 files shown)


test/functional/host/subscription/content_override_test.rb
```diff
@@ -47,6 +47,36 @@
...
+    end
+    ex.returns({})
+
+    expected_result = success_result("Updated content override.\n")
+
+    result = run_cmd(@cmd + params)
+    assert_cmd(expected_result, result)
+  end
+
+  it "attaches a content override with value other than enabled without --force" do
```
jeremylenz, line 70:
```suggestion
  it "does not attach a content override with value other than enabled without --force" do
```

jeremylenz (commented):
thanks @chris1984 !

just one more :)

test/functional/host/subscription/content_override_test.rb
chris1984: Thanks for catching it :) Updated.

jeremylenz (approved):

PR #979, opened 2024-12-17
Refs #37151 - Fix --force flag to use option_override_name

jeremylenz:
https://github.com/Katello/hammer-cli-katello/pull/977 should have validated `option_override_name`, but inadvertently looked at `option_value` instead. This caused `hammer content-override` to break completely.

Turns out our Hammer tests were also wrong. Params should be like
```
--override-name enabled
--value true

--override-name enabled
--value 1
```

but instead were like
```
--value enabled
```

To test:

```
hammer activation-key content-override --id 2 --content-label satellite-client-6-for-rhel-9-x86_64-rpms --value true
hammer activation-key content-override --id 2 --content-label satellite-client-6-for-rhel-9-x86_64-rpms --override-name foo --value 0
```

1. Without the `--override-name` flag, values `true`, `false`, `0`, and `1` (and really, any value) should work.
2. With the `--override-name` of `enabled`, the values above should also work.
3. With the `--override-name` of anything other than `enabled`, you should get the (new & improved) error
```
Could not update content override:
  You must use --force to set an override other than 'enabled'
```

### DIFF
--- a/lib/hammer_cli_katello/content_override.rb
+++ b/lib/hammer_cli_katello/content_override.rb
@@ -31,7 +31,7 @@ module HammerCLIKatello
           if option(:option_remove).exist?
             option(:option_value).rejected
           elsif option(:option_value).exist?
-            if !@option_values['option_value'].casecmp('enabled').zero? &&
+            if !@option_values['option_override_name'].casecmp('enabled').zero? &&
                @option_values['option_force'] == false
               raise ArgumentError, _("You must use --force to set a value other than 'enabled'")
             end
--- a/lib/hammer_cli_katello/content_override.rb
+++ b/lib/hammer_cli_katello/content_override.rb
@@ -33,7 +33,7 @@ module HammerCLIKatello
           elsif option(:option_value).exist?
             if !@option_values['option_override_name'].casecmp('enabled').zero? &&
                @option_values['option_force'] == false
-              raise ArgumentError, _("You must use --force to set a value other than 'enabled'")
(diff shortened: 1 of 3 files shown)


chris1984 (approved):
Code looks great, ACK

Before PR:
```bash
[root@ip-10-0-167-154 ~]# hammer activation-key content-override --id 1 --content-label satellite-client-6-for-rhel-9-x8
6_64-rpms --value true
Could not update content override:
  You must use --force to set a value other than 'enabled'
  
  [root@ip-10-0-167-154 ~]# hammer activation-key content-override --id 1 --content-label satellite-client-6-for-rhel-9-x8
6_64-rpms --override-name foo --value 0
Could not update content override:
  You must use --force to set a value other than 'enabled'
```

After PR:
```bash
[root@ip-10-0-167-154 ~]# hammer activation-key content-override --id 1 --content-label satellite-client-6-for-rhel-9-x86_64-rpms --value true
Updated content override.
[root@ip-10-0-167-154 ~]# hammer activation-key content-override --id 1 --content-label satellite-client-6-for-rhel-9-x86_64-rpms --override-name foo --value 0
Could not update content override:
  You must use --force to set an override other than 'enabled'
  [root@ip-10-0-167-154 ~]# hammer activation-key content-override --id 1 --content-label satellite-client-6-for-rhel-9-x86_64-rpms --override-name enabled --value 0
Updated content override.

[root@ip-10-0-167-154 ~]# hammer activation-key content-override --id 1 --content-label satellite-client-6-for-rhel-9-x86_64-rpms --override-name foo --value 0 --force
Updated content override.
```

PR #982, opened 2025-02-05
Added a hammer devel container script

parthaa:
Bash script to turn on the hammer devel environment.
Also updated readme.


chris1984 (approved):
LGTM

PR #988, opened 2025-03-05
Use trusted publishers to release the gem

ekohl:


chris1984 (approved):

ekohl:
I created the trusted publisher and removed the old API key, but I can't clean up the old GH secret since I'm not an admin on Katello.

chris1984:
@ianballou or @parthaa can you clean up the secret?

PR #993, opened 2025-04-28
Fixes #38392 - show container upstream name in repository list

ianballou:
Adds container upstream name optionally to the repository list command. Try --fields=ALL to see it.


sjha4 (approved):
Ack..Able to see new field in hammer.
Screenshot from 2025-05-07 13-54-32

PR #998, opened 2025-07-22
Fixes #38612 - Remove rolling CV from host info command

qcjames53:
Hammer-cli-katello 1.17 exposes the "rolling CV" fields on Foreman 3.14 despite rolling CV not being "released" until 3.15+. We need to patch 1.17 to remove these fields from the default host info view.

**How to test**

1. Pull the 1.17.z branch of hammer-cli-katello and make sure it runs with the appropriately-outdated version of hammer-cli and hammer-cli-foreman (grab versions from foreman 3.14ish).
2. Run a `hammer host info` command and see that "Rolling CV" is not displayed next to "Composite CV".
3. Run a `hammer host info` command with `--fields ALL` and see that "Content view" -> "Rolling" is not displayed next to "Composite".
4. Ensure the rolling CV fields are not exposed when manually triggering them via `fields "Content Information/Content view environments/Rolling CV"`

### DIFF
--- a/lib/hammer_cli_katello/host_extensions.rb
+++ b/lib/hammer_cli_katello/host_extensions.rb
@@ -66,14 +66,11 @@ module HammerCLIKatello
                     :replaced_by => [_('Content Information'), _('Content View Environments'), _('CV Name')].join('/')
                   field :composite, _("Composite"), Fields::Boolean,
                     :replaced_by => [_('Content Information'), _('Content View Environments'), _('Composite CV')].join('/')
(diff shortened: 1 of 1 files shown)


qcjames53:
I believe the non-blocking single commit test is failing because this is merging into 1.17.z instead of main. It can be safely ignored.

chris1984 (approved):
Looks fine to me, are we shipping it with 6.18 or are we going to include it in 6.19?

PR #1026, opened 2026-04-16
Remove Katello Agent and Katello Events

jturel:


chris1984 (commented):
I think this can be taken out of draft if I remember right.

jturel:
> I think this can be taken out of draft if I remember right.

I wish! This PR was a little early - it depends on some bigger pieces of work which are essentially experiments right now:
- https://github.com/Dynflow/dynflow/pull/468
- https://github.com/jturel/katello/pull/1

PR #1031, opened 2026-05-21
Keep backwards compatibility with hammer_cli_foreman 3.x

zjhuntin:
## Summary
- Changes `hammer_cli_foreman` dependency from `~> 5.0` (requires 5.0+) to `>= 3.10, < 6.0`
- Allows the plugin to work with both 3.x and 5.x releases
- Follow-up to #1030 based on reviewer feedback that plugins don't need 5.0.0 specifically

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/hammer_cli_katello.gemspec
+++ b/hammer_cli_katello.gemspec
@@ -68,6 +68,6 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency 'syslog', '~> 0.1.2'
   gem.add_development_dependency 'theforeman-rubocop', '~> 0.1.0'
 
-  gem.add_dependency 'hammer_cli_foreman', '~> 5.0'
+  gem.add_dependency 'hammer_cli_foreman', '>= 3.18', '< 6.0'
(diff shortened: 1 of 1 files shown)


hammer_cli_katello.gemspec
```diff
@@ -68,6 +68,6 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency 'syslog', '~> 0.1.2'
   gem.add_development_dependency 'theforeman-rubocop', '~> 0.1.0'
 
-  gem.add_dependency 'hammer_cli_foreman', '~> 5.0'
+  gem.add_dependency 'hammer_cli_foreman', '>= 3.10', '< 6.0'
```
ogajduse: I'd make it stricter and set `>= 3.18`. To restrict 3.18 and older version users from using hammer-cli-katello 5.0, but make sure 3.19 users can still use it, as we don't anticipate any breaking changes to come in 5.0.

chris1984 (approved):

